PR #647, opened 2019-01-04
birdseye integration

alexmojaki:
This is a stab at integrating birdseye into Thonny as discussed in the mailing list. For now it’s very preliminary and full of copy pasted code. I wanted to have something to directly play with to see what it’s like to use and to get a feel for the coding challenges. I hope you’re OK with having a discussion here instead of in a plain issue. For now don’t fuss about the code.

I suggest checking out this branch and giving the integration a try. Just click “Run > Debug current script (birdseye)”. It automatically runs the server and opens the result in the browser.


aivarannamaa:
Sorry for the delay!

Unfortunately I didn't get it working in Windows. I installed "birdseye" and "requests" (required by "outdated") via Tools => Manage packages.

I tried it with following script:

```
from birdseye import eye

@eye
def foo():
    print(2 + 22)

foo()
```

When I select "Run => Debug current script (birdseye)", I get 

```
>>> %Birdseye bird.py
PROBLEM WITH THONNY'S BACK-END:

Traceback (most recent call last):
  File "C:\workspaces\python_stuff\thonny\thonny\backend.py", line 205, in handle_command
    response = handler(cmd)
  File "C:\workspaces\python_stuff\thonny\thonny\backend.py", line 375, in _cmd_Birdseye
    return self._execute_file(cmd, BirdsEyeRunner)
  File "C:\workspaces\python_stuff\thonny\thonny\backend.py", line 803, in _execute_file
    self._ast_postprocessors
  File "C:\workspaces\python_stuff\thonny\thonny\backend.py", line 816, in _execute_source
    source, filename, execution_mode, ast_postprocessors
  File "C:\workspaces\python_stuff\thonny\thonny\backend.py", line 1245, in execute_source
    webbrowser.open_new_tab('http://localhost:7777/ipython_call/' + eye._last_call_id)
TypeError: can only concatenate str (not "NoneType") to str
```

alexmojaki:
Hi! Thanks for getting to this.

I can't reproduce this, which is worrying. I also can't think of why Windows are any other differences in your setup could cause a problem, let alone the one you're seeing.

It looks like the code didn't execute, at least not under birdseye. It doesn't seem to have printed out `24`. Are there any calls showing if you visit localhost:7777? Can you step through the code starting from

    result = self._prepare_hooks_and_execute(source, filename, global_vars)

and see what happens, at least on the thonny end? 

The point of the integration is that you don't need `@eye`, the command takes care of the tracing. What happens if you remove that? Similarly, what if you run the current script normally?

Can you try running different scripts with the birdseye command and see if any of them work? Try at least one with no functions.

Does using a different filename make a difference?

The traceback suggests that you weren't running the exact code that I had in my branch, maybe that caused the problem? Please checkout this branch exactly. I've also just fixed a minor problem I just noticed.

aivarannamaa:
Together with last commit I got it working, both with and without the @eye decorator.

Have you considered reusing the browser tab for each debug run?

I mostly understand what your code aims to do, but what is the purpose of following lines:

```
        sys.stderr = stream_proxy(sys.stderr)
        sys.stdout = stream_proxy(sys.stdout)

```
?

alexmojaki:
Awesome!

> Have you considered reusing the browser tab for each debug run?

Seems like a good idea, but how? At least on my setup `webbrowser.open` doesn't seem to allow that. Can it do that on yours? Do you want some external package just for that?

The stream proxy stuff hides all output from the server. I use the same code in the Jupyter/IPython integration, it seems to work fine. But it's a hack, and I wouldn't be surprised if you could come up with a better idea. At least one possibility is to isolate the server completely by starting it with `subprocess`, although this may introduce new complications.

aivarannamaa:
Looks like I didn't remember correctly how `webrowser.open` works, nevermind ...

I tried if I can simplify your code and package it as a plug-in and I think I succeeded:
[thonny-birdseye.zip](https://github.com/thonny/thonny/files/2761628/thonny-birdseye.zip) (NB! Needs latest version of the master).
See https://github.com/thonny/thonny/wiki/Plugins about the plug-in format.

One remaining issue is how to avoid launching new birdseye server if it is already launched. Can you change your `birdseye` command so that it gives a warning and exits when the server is already running?

alexmojaki:
Technically, it does show a message and exit when the server is already running, it's just that it includes a long traceback. Do you want less noise in logs? It's possible to hide all output:

```
    subprocess.Popen(
        [sys.executable, "-m", "birdseye"],
        stderr=subprocess.DEVNULL,
        stdout=subprocess.DEVNULL,
    )
```

but I guess you might want to see some logs for debugging.

On a side note, the current approach leaves birdseye running when Thonny is closed. This bothers me a little but I can't see any serious problem with it. What do you think?

Are you hoping that this will be a separately maintained and packaged plugin? There's still a decent chunk of duplicated logic such as:

```
            return {"context_info": "after normal execution"}
        except Exception:
            return {"user_exception": self._vm._prepare_user_exception()}
        finally:
            del sys.meta_path[0]
            if hasattr(self._vm, "_original_import"):
                self._vm._restore_original_import()
```

and that does seem like a serious problem if this is going to live in a different codebase. Do you think you can provide an API that takes care of that stuff?

aivarannamaa:
> Technically, it does show a message and exit when the server is already running, it's just that it includes a long traceback.

I'll test it on Linux later, but at least on Windows it happily started new Python process without any stacktrace or notification about server already running and the process didn't resume at least in a minute.

> Are you hoping that this will be a separately maintained and packaged plugin?

Yes. I don't want to include it in the main project mostly because of many (transitive) dependencies. 

At the moment there is no clear line between API and internal stuff. I'll go over relevant parts of `thonny.backend` module so that the plug-in doesn't need to depend on underscored names.

If you don't want to manage the plugin then I can do it myself.

alexmojaki:
Oh right, I vaguely remember now that Windows allows multiple processes listening on the same port. I still don't get how that works out in practice. Unix immediately gives an error `OSError: [Errno 48] Address already in use`. This isn't due to any code that I wrote.

> the process didn't resume at least in a minute.

I don't follow. Which process? I would expect everything to continue normally: the browser opens a tab which tries to reach localhost:7777 and at least one of the started server processes responds, even if there are many or some die.

> I don't want to include it in the main project mostly because of many (transitive) dependencies.

What if clicking the debug with birdseye button prompts installing birdseye with pip if it's not already installed (i.e. importing fails)?

Requiring users to separately discover the existence of and install birdseye means far fewer users will install it. Now of course, I don't want that to happen, but I'm very biased. What's your honest opinion on the value of birdseye as an alternative debugger?

> At the moment there is no clear line between API and internal stuff. I'll go over relevant parts of thonny.backend module so that the plug-in doesn't need to depend on underscored names.
>
> If you don't want to manage the plugin then I can do it myself.

I'm not concerned about using underscored names or managing the plugin. I just don't want a change to be made in the main repo and then nobody remembers that the same change needs to be made in the plugin.

aivarannamaa:
> the process didn't resume at least in a minute.

I meant the second instance of the sever seemed to remain running, instead of delaying the error message.

> What if clicking the debug with birdseye button prompts installing birdseye with pip if it's not already installed (i.e. importing fails)?

Good idea! In this case I'm happy to include the plugin in the main codebase. If the command handler can't import birdseye, then it shows a messagebox which gives a short description of the feature, instructions for installation (via Tools => Manage Plug-ins) and allows user to visit https://birdseye.readthedocs.io for more information. I can also give an introduction and links on a separate help page.

> What's your honest opinion on the value of birdseye as an alternative debugger?

If I was still teaching, I would mostly use the "nicer" debugger (because I feel it is more direct and understandable and because I'm biased :) ), but I would also show the birdseye visualization to offer alternative view to the same thing. I like how birdseye presents the nesting in the code structure and execution states. I believe some teachers may prefer it to the unidirectional "nice" debugger.

aivarannamaa:
About the server -- can you change the implementation so that it won't start second instance in Windows?

aivarannamaa:
> About the server -- can you change the implementation so that it won't start second instance in Windows?

Nevermind, I'll better run the server via api in a separate thread in frontend.

alexmojaki:
OK, so what's the current state of things? Is there anything you'd like me to do?

aivarannamaa:
I integrated current code and added help page and instructions dialog (#660). I'll refactor the server stuff later.

aivarannamaa:
Implemented under #660

PR #663, opened 2019-01-23
Birdseye refactor

alexmojaki:
This mainly removes duplicated code between birdseye_backend.py and backend.py to make maintenance easier and less error-prone.


aivarannamaa:
Thanks!

PR #667, opened 2019-01-24
Single step execution

cubimon:
This is not a full implementation. I just added some things, that made the current implementation less useable.

- refocus code editor on error/output
- move to next line of code, which is probably the better default behaviour than staying on the same line → continuously pressing F9 is "like" debugging
- fix a final missing endline if a whole function definition was selected and executed with F9. This is kinda meh atm, because you have to select the whole definition and stepping with F9 doesn't work for now
- Control+q is used by lots of linux gui application to terminate them, e.g. firefox/gitk


cubimon:
I just tested some applications:

Alt+F4 works:

- chromium

Ctrl+Q works:

- gitk
- Firefox
- Pavucontrol
- Discord
- Vlc
- Visual Studio Code (vim extension must be disabled)
- Thunderbird
- Kile
- JDownloader
- Libre Office
- Etcher
- Simple Scan
- Teamspeak
- Wireshark
- Processing IDE
- Calibre
- QtCreator
- Audacity

neither works:

- Steam
- mpv
- vim/nvim (obviously ¯\\\_(ツ)_/¯)
- Konsole/probably all terminals
- Eclipse
- Android Studio

But I see your issue, both shortcuts should work on Linux

aivarannamaa:
Hmm, from this it looks like replacing Alt+F4 with Ctrl+Q makes sence. I can later add Alt+F4 as extra shortcut, which is not displayed in the menu.

Which distro are you using?

cubimon:
I use Arch with i3wm. You should definitely readd alt+f4, because your users may be used to it

PR #671, opened 2019-01-27
Use Path() and path_prefix

fstengel:
Use the Path() object
Use the "path_prefix" option

Thus, if a config page (in the current case the ESP32 config page) sets path_prefix it will be used for main.py and boot.py

### DIFF
--- a/thonny/plugins/micropython/__init__.py
+++ b/thonny/plugins/micropython/__init__.py
@@ -815,27 +815,19 @@ class MicroPythonProxy(BackendProxy):
     def _cmd_upload(self, cmd):
         # Target is interpreted relative to the root
         if len(cmd.args) == 1:
-            source = cmd.args[0]
+            source = Path(cmd.args[0])
             # target is at root
-            target = os.path.basename(source)
+            target = source.name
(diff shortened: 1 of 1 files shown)


aivarannamaa:
Sorry, I didn't understand. Does your PR fix a bug or enhances something?

fstengel:
It does enhance things by giving a prefix/base path option for the ESP32 backend. This permits me to use the lobo port and tell the backend that the internal file system is rooted at `/flash/` rather than `/`
This enhancement doesn't do anything without the corresponding part in thonny/thonny/esp...
the modification in thonny-esp adds a text entry to set/edit the option that is used by the micropython backend. If the option isn't set, then it defaults to the empty string.

aivarannamaa:
Thank you for the idea!

I'll add this path_prefix method to the base plugin and introduce separate `ESP32 (LoBo)` backend in `thonny-esp` plug-in, which overrides the path-prefix. This way user only needs to select proper back-end and doesn't need to worry about path prefix.

I don't see the need for replacing path strings with `Path` objects, though.

PR #673, opened 2019-01-30
Simplify and canonicalise setup.py.

Natureshadow:


Natureshadow:
This was a bad idea.

(But anyway, reading requirements.txt in setup.py generally is.)

Issue #692, opened 2019-02-13
File save cancel generates an error.


peterdcasey:
When running / saving a new file <untitled> and selecting Cancel, an error window pops up with the error below. (Windows v3.1.2>
=================================================
Traceback (most recent call last):
  File "C:\Users\pcasey\AppData\Local\Programs\Thonny\lib\tkinter\__init__.py", line 1705, in __call__
    return self.func(*args)
  File "C:\Users\pcasey\AppData\Local\Programs\Thonny\lib\site-packages\thonny\running.py", line 317, in _cmd_run_current_script
    self.execute_current("Run")
  File "C:\Users\pcasey\AppData\Local\Programs\Thonny\lib\site-packages\thonny\running.py", line 282, in execute_current
    script_dir = normpath_with_actual_case(os.path.dirname(filename))
  File "C:\Users\pcasey\AppData\Local\Programs\Thonny\lib\ntpath.py", line 221, in dirname
    return split(p)[0]
  File "C:\Users\pcasey\AppData\Local\Programs\Thonny\lib\ntpath.py", line 183, in split
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not NoneType

aivarannamaa:
Fixed in #714

Issue #691, opened 2019-02-13
No Show shell on run with no input() prompt


peterdcasey:
Thank you so much for showing the Shell window when code is run, v3.1.2. One small issue is it appears the Shell does not show if there is not output to the Shell. Using an input() with no prompt as the first statement in the code does not reveal the Shell.

Thanks again for the update.

aivarannamaa:
Fixed in #714

Issue #630, opened 2018-12-17
Pressing up then down in shell doesn't leave shell in previous state


aivarannamaa:
**[Original report](https://bitbucket.org/bitbucket-issue-migration\thonny-issues.zip/issue/630) by Alex Hall (Bitbucket: [Alex_Hall](https://bitbucket.org/Alex_Hall)
 ).**

----------------------------------------

In the shell, if I press up three times, it will go through my previous three commands - great. Pressing down usually undoes this by going back to more recent commands - great. The third down, however, should leave the shell in the state it was before all this, having undone all three ups. Instead the previous command just sits there.

This can be mildly inconvenient if your shell was blank before and you now have to manually delete the previous command. It can be really annoying if you were typing some code which is now lost.

aivarannamaa:
> It can be really annoying if you were typing some code which is now lost.

This part is not fixed yet.

PR #714, opened 2019-03-24
fix for issues #630 #691 and #692

chpurdy:
The cause for the error crashing when canceling the save dialog when running a new program was that there is no check that the filename was actually assigned.


aivarannamaa:
Sorry for the delay in responding and thanks a lot for this PR!

PR #725, opened 2019-04-07
When saving a file, respect the original fileformat.

badukaire:
This is done by:

- capturing the fileformat with attribute newlines
- get_content_in_bytes optionally provides windows format

### DIFF
--- a/thonny/code.py
+++ b/thonny/code.py
@@ -48,6 +48,7 @@ class Editor(ttk.Frame):
         self.columnconfigure(0, weight=1)
         self.rowconfigure(0, weight=1)
 
+        self._newlines = None
         self._filename = None
         self._last_known_mtime = None
         self._asking_about_external_change = False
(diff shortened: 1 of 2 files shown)


thonny/code.py
```diff
@@ -157,6 +158,9 @@ def _load_file(self, filename, keep_undo=False):
         with tokenize.open(filename) as fp:  # TODO: support also text files
             source = fp.read()
 
+        # get the file file format (DOS/UNIX)
+        self._newlines = fp.newlines
```
aivarannamaa: Good find! I did not know about this attribute. 

I would expect this to be under `with`, but apparently closing the file pointer (which happens after exiting `with` block), leaves `newlines` property intact.

thonny/code.py
```diff
@@ -241,7 +245,7 @@ def save_file(self, ask_filename=False):
             self._filename = new_filename
             get_workbench().event_generate("SaveAs", editor=self, filename=new_filename)
 
-        content = self._code_view.get_content_as_bytes()
+        content = self._code_view.get_content_as_bytes(self._newlines == '\r\n')
```
aivarannamaa: Did you know that `filenames` attribute can be a tuple? (https://stackoverflow.com/a/29832736/261181)

Anyway, I think it is good as it is now, because it makes sense to use platform format if the original file contained more than one type of linebreak.

aivarannamaa:
Thanks! Good job!

PR #727, opened 2019-04-08
respect fileformat for unix files too

badukaire:
There is another issue with file formats. When editing on windows files with unix fileformat, they were overwritten with DOS fileformat. This should not happen neither.

For fixing that:
- on `get_content_as_bytes`, pass the newline format without comparing it (it's `None` by default)
- modify the condition to save on windows file format

I have tested this both on windows and linux machines.

Let me know if you have comments regarding coding style etc.

### DIFF
--- a/thonny/code.py
+++ b/thonny/code.py
@@ -245,7 +245,7 @@ class Editor(ttk.Frame):
             self._filename = new_filename
             get_workbench().event_generate("SaveAs", editor=self, filename=new_filename)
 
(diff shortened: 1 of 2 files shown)


aivarannamaa:
You are right, the condition was too lax and we were both hasty :)

badukaire:
Excuse me, I am not very familiar with the github pull request. I'd swear I had created a draft code request. Am I wrong?

There is a small thing to be uploaded, as it is it may be slightly wrong.

PR #728, opened 2019-04-08
default value of None

badukaire:
There is a call to get_content_as_bytes without a parameter, therefore the default value of None should be set. Otherwise there would be a runtime error.

Sorry for so many pull requests, I thought my pull request was a draft.


aivarannamaa:
It was my fault actually. It was a draft PR, but after I clicked "start review" button (or something like this), it changed. So, the problem was that I am not too familiar with GitHub pull requests :)

Issue #668, opened 2019-01-25
Translation: PR welcome?


cspaier:
Hi,

A group of teachers (me included) are using Thonny in a highschool in south France. We distribute a portable version of Thonny [here](https://github.com/cspaier/pydiderot) with small customization and homemade libraries for students and teachers.

I noticed your [faq](https://bitbucket.org/plas/thonny/wiki/FAQ) related to translating thonny. I was contemplating marking the strings with [gettext](https://docs.python.org/3/library/gettext.html) so we could render a po file and host translation on externate tools such as [poeditor](https://poeditor.com/).

Before starting working on that, I wanted to check with you if you would welcome such a PR.

Also maybe there are some specific technical details I missed on this matter?

Anyway thanks for good job! Thonny is a great tool for teaching and learning python.

aivarannamaa:
Hi!

Thank you for the kind words!

Internationalization still isn't a priority for me, but it looks like there are enough people who consider it important. Therefore I'll welcome your PR (although this means I'll need to be more careful in changing strings in the future).

Translating short and simple UI texts (eg. menu items) should go smoothly. Longer texts (eg. Assistant's error explanations) need more care so that a small change in the original text wouldn't change all translations unusable. I hope there are features for this in gettext (eg. by giving codenames for some strings).

cspaier:
Thank you for your answer.

Hum, I think gettext will handle that smoothly. If you change a string, we will have to render new po file and update translations. In the meantime, the untranslated strings will be shown in English so nothing will be break.

I fear I don't really understand the "codenames" idea you are referring too.

Anyway the more important might be the easy "menu strings" since that's what the students will be using mostly.

Maybe I start working on a PR and we could discuss specifics there if needed?

aivarannamaa:
> I fear I don't really understand the "codenames" idea you are referring too.

I thought of something like `_("Blah, blah, ...", id="blah_msg")` so that the translations would be bound to `blah_msg` and this allows changing `"Blah, blah, ..."` without breaking the translations. But nevermind, looks like (Python's) gettext doesn't support this.

> Maybe I start working on a PR and we could discuss specifics there if needed?

Sure. I wish you luck!

cspaier:
I think `gettext` handle that. It will bind the strings with it's origins (`file.py line 13`).

Anyway, I will do some tests and let you know.

badukaire:
do you have any news regarding this? do you have some draft code around to see?

cspaier:
[...]

PR #736, opened 2019-04-18
Internationalisation (i18n)

cspaier:
This PR activate i18n functionalities for thonny using python [gettext](https://docs.python.org/3/library/gettext.html)  and the [babel](http://babel.pocoo.org/en/latest/) library.

As explained in the related issue (#668 ), and in particular [this comment](https://github.com/thonny/thonny/issues/668#issuecomment-482491775) the changes are:

- A new `thonny/languages.py` file containing a dict of language we use.
- Two new `workbench` methods to set up and initialize language at startup from configuration.
- A new `language` plugin that allow user to set up their language.
- A folder `thonny/locale` containing the messages catalogs.
- A `docs/translatemd` file explaining how to work with those catalogs files using the babel library.

I marked some strings but I am positive I did NOT got them all. It's a boring a repetitive work and if anyone feel like marking the rest he is welcome.

I will discuss in the related issue the translation organization.


aivarannamaa:
Thank you for your work!

PR #744, opened 2019-04-28
Added some strings to French translation

XanderLeaDaren:


aivarannamaa:
I assume same content was overridden in #745

PR #752, opened 2019-05-05
Page down to end of shell text when focusing

### LINKED ISSUE #749: Focus shell option should bring you to a new prompt
Currently it just takes you back to where the editor cursor was in the shell.

However if I would like to efficiently flip between editing and running expressions, I would like to be able to press alt+S and immediately start typing a new expression to be evaluated.

Or optionally/alternately, be taken to the end of textline for a previously-typed but yet-to-be-evaluated expression, e.g

```
>>> a = [x[0] for x in some_script_function(argu
```
be brought to the end of the u, so I can finish writing it and hit enter (or hit ctrl+shift+left and delete everything to write a new expression).

If I had previously selected something to copy, or scrolled around some output, I don't want to be brought back to there with alt+S.

Ivoz:
Fixes #749

new at "Tkintering", I hope this is enough behaviour to work correctly. Unless you need something to add a prompt in some situations? 🤔 Not sure.

### DIFF
--- a/thonny/workbench.py
+++ b/thonny/workbench.py
@@ -1812,6 +1812,10 @@ class Workbench(tk.Tk):
 
     def _cmd_focus_shell(self) -> None:
         self.show_view("ShellView", True)
+        shell = self.get_view("ShellView")
+        # go to the end of any current input
+        shell.text.mark_set("insert", "end")
+        shell.text.see("insert")
 
     def _cmd_toggle_full_screen(self) -> None:
         """


aivarannamaa:
If you just want to put entry cursor to the end of the text and scroll into this location, then standard way to do this would be:

```
text.mark_set("insert", "end")
text.see("insert")
```

Would this work for your goal as well?

Ivoz:
Sure. I was wondering if `perform_page_down` had any good error checking worth incorporating into that. If you think simpler is better than I won't disagree

Ivoz:
Updated commit to reflect that

Ivoz:
Also wondering if it would be handy to have a method to get the current shell in `Workbench`, similar to `self.get_editor_notebook()` / `self._editor_notebook` ? Or is it not there because shell could be "optional" and not shown / not instantiated?

aivarannamaa:
> Also wondering if it would be handy to have a method to get the current shell in Workbench, similar to self.get_editor_notebook() / self._editor_notebook ? Or is it not there because shell could be "optional" and not shown / not instantiated?

It would be nice to have `get_shell` method, actually. At some point in time it would be too early to call it, but same applies to other methods as well. It's OK, to get and create the shell even if user has hidden it -- it just won't be shown then.

aivarannamaa:
Thanks!

I forgot that there is `get_shell` in `thonny` module. I changed the code to use it.

PR #773, opened 2019-06-08
Chinese (zh-TW) translation

ShenTengTu:
I found there are not Traditional Chinese (zh-TW) translation, so I work it .

I see POEditer has "Chinese" option which is only for Simplified Chinese, so I suggest "Chinese (zh-TW)" option which is for Traditional Chinese in Taiwan.

I cureently translate according to the situation I can understand.

Edit:
I found some files not pass `back check`  but they are not my work part so I keep it temporarily.


aivarannamaa:
Thank you!
Sorry about the black errors in my own code.

Dleta:
Brother,we have give up Tranditional Chinese yet.it is so complex.and I suggest you learn and use Simplified 
Chinese .It is much easier than Tranditional Chinese.@ShenTengTu

PR #775, opened 2019-06-09
French help

georgesk:
Hello, here is the translation of help pages to French.
Files `thonny/plugins/help/__init__.py` and `thonny/workbench.py` bear some modifications which are the same as in my other pull request, for consistency of the present pull request.

Thank you for considering this pull request shortly, as it does not *change* much parts in your master branch. Most of this pull is made of additions.

About the creation of a subdirectory en_US : please feel free to create one such directory. Language purists may suggest creating also us_GB, en_IE, en_IN, etc. However I do believe that default help files should still remain in the parent directory (if they were written and checked by a native speaker, then they can be copied into some en_FOO subdirectory right now).

### DIFF
--- a/thonny/plugins/help/__init__.py
+++ b/thonny/plugins/help/__init__.py
@@ -3,10 +3,10 @@ import tkinter as tk
 import tkinter.font
 from tkinter import ttk
 
-from thonny import get_workbench, rst_utils, tktextext, ui_utils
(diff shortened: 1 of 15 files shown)


aivarannamaa:
Thank you!

I added some fixes in https://github.com/thonny/thonny/commit/245e888a57f17c8f4aa0639a74421a3c7c6dc34f. Some rst files had syntax errors in them (diagnostic messages are printed to stderr/stdout when you open help page with syntax problems).

PR #791, opened 2019-06-29
Update German translation

be-we:


aivarannamaa:
Thank you!

aivarannamaa:
Are your translations here as well: https://poeditor.com/projects/po_edit?id_language=55&id=256075 ?

If not then it's recommended to import them there as well, otherwise next update may override yours.

be-we:
@aivarannamaa The given link only leads to an error 404.

aivarannamaa:
I see... Probably because you haven't logged in. You should join Thonny project on Poedit as described here: https://github.com/thonny/thonny/blob/master/docs/translate.md

be-we:
@aivarannamaa Joining the Thonny project on POEditor has solved the error 404. Then I entered my translation updates into POEditor as well.

aivarannamaa:
Great, thank you!

Hermann12:
I have updated the translation and tested the .mo and  .po within thonny. Some translations are not taken? e.g. "Run current script" isn't into the .mo file why? I searche in the POEditor, but coudn't find. Also other menue text.

PR #842, opened 2019-08-09
completed translation

Dleta:
I completed translation too(Simplified Chinese).


aivarannamaa:
Thanks!

I added back the [BETA] label. I'd like to keep it for some time, because I can't review the translation myself.

PR #848, opened 2019-08-12
fix & update zh_TW translation (3.2.0rc1)

ShenTengTu:
Update zh_TW translation (3.2.0rc1)
- File browser commands & more
- Fix some words that were not displayed correctly before


aivarannamaa:
Thanks!

PR #857, opened 2019-08-15
Use UNIX sockets for IPC. Closes #854.

### LINKED ISSUE #854: Unable to use and information leak on multi-user systems
We tried using Thonny in an educational environment with a remote desktop server used by multiple users in parallel, to find out that this is jut not possible currently.

When starting Thonny, it will find the open TCP socket of the instance run by another user, and connect to it.

This causes THonny to not start for more than one user, and also leaks any action triggered on the command-line to the instance of the other user.

Please use UNIX domain sockets in a private directory of the user running thonny instead, and only fall back to TCP on WIndows (if you really need to support this platform).

Natureshadow:
I tested this on Linux, someone will still have to test on WIndows.

While implementing, I stumbled upon another bug which I will report now.

### DIFF
--- a/misc/sockatse.py
+++ b/misc/sockatse.py
@@ -1,16 +1,33 @@
+from getpass import getuser
+import os
 import socket
 from time import sleep
 
-port = 4957
+from xdg import XDG_RUNTIME_DIR
+
+def get_socket_path():
+    base = XDG_RUNTIME_DIR or os.environ.get("TMPDIR", None) or  os.environ.get("appdata", None) or "/tmp"
+    path = os.path.join(base, "thonny-%s" % getuser())
+
+    try:
+        os.mkdir(path)
+    except FileExistsError:
+        pass
+
+    if not os.name == 'nt':
+        os.chmod(path, 0o700)
+
+    return os.path.join(path, "ipc.sock")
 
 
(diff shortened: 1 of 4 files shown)


Natureshadow:
The test suite fails because it wants to reformat lines I did not touch.

aivarannamaa:
Thanks for ideas!

It looks like AF_UNIX is not supported by Windows version of Python yet (https://bugs.python.org/issue33408).

Is there something specific about `$XDG_RUNTIME_DIR`? If not then I'd rather use `thonny.THONNY_USER_DIR` as socket directory.

Natureshadow:
>Is there something specific about `$XDG_RUNTIME_DIR`? If not then I'd
>rather use `thonny.THONNY_USER_DIR` as socket directory.

It is not guaranteed that /home is on a filesystem that supports sockets (e.g. sshfs or CIFS). /run, however, is there for exactly that purpose.

aivarannamaa:
Implemented in 2c82d84ac0d9dba019e0bae7ca6df2592e97ab3a and 9e51e8e82317e0d3760fad9632ba7f8ec80f61a4

Natureshadow:
Thank you that you did it, after that long discussion!

PR #860, opened 2019-08-15
Use xdg module to only keep XDG spec in one place.

Natureshadow:


aivarannamaa:
I prefer not to add another dependency for so simple tasks

Natureshadow:
>I prefer not to add another dependency for so simple tasks


Good point, but reimplementing standards is hard. That's why you forgot to fall back to the XDG standard's default for RUNTIME_DIR, and that in turn is why I like keeping standard adherence in one place :).

aivarannamaa:
> you forgot to fall back to the XDG standard's default for RUNTIME_DIR

I don't see any fallbacks for this in xdg module either: https://github.com/srstevenson/xdg/blob/master/src/xdg/__init__.py#L120

PR #861, opened 2019-08-16
Adds European Portuguese translation (pt_PT)

emansije:
Also marks for translation some strings that were still unmarked.


aivarannamaa:
Thanks for the translation, but I'd rather mark the strings myself. I prefer to keep some strings untranslated for now in order to reduce translation burden (for all languages). In some places I still may change the original strings.

I'll reject this PR, but I'll include your translations from POEditor.

Thanks again!

aivarannamaa:
I now noticed that translation docs kind of encouraged marking new strings. I updated this (https://github.com/thonny/thonny/commit/45ee6e7cb774e7f7969b2f9ba5215a52cbec23ee). Sorry about the confusion!

emansije:
I was, indeed, following the instructions from the documentation. It's OK, I'll just keep an eye out for updates on POEdit.

How do you want me to suggest the marking of new strings? Can you pick the ones you can use from my PR?

I've also noticed that I mistakenly included commit 9c23f81 ("Prevents internal error in Object Inspector") in this PR instead of making a new one. I stumbled on the issue while testing the translations, so I did a quick fix. I'll just make a new PR for this issue, then.

aivarannamaa:
> I've also noticed that I mistakenly included commit 9c23f81 ("Prevents internal error in Object Inspector") in this PR instead of making a new one. I stumbled on the issue while testing the translations, so I did a quick fix. I'll just make a new PR for this issue, then.

Please do!

> You seem to have forgotten to add a new line with:
> "pt_PT": "Português (PT) [BETA]",
> for European Portuguese.

Oops, fixed now

PR #864, opened 2019-08-18
Prevents internal error in Object Inspector

emansije:
When switching between `Data` and `Atts` twice, the `object_info` dictionary would throw
an internal error because it didn't have the `attributes` key.


aivarannamaa:
Thanks!

PR #934, opened 2019-10-08
update translation

rnLIKEm:
Hi,I noticed an issue,the width is not enough to show Chinese characters but enough in English.
By the way,I update these several words for more precise.


aivarannamaa:
Thanks! Did you also update the language in Poeditor.com?

I added padding to button captions, so they should be visible now

rnLIKEm:
Yes, I updated it.

PR #981, opened 2019-11-10
Added Polish translation

jmiszczak:
Almost compete Polish translation, created using POEditor.


aivarannamaa:
Thank you! The translation will be included in the next release.

PR #991, opened 2019-11-15
fixes #987 micropython autocomplete issues

### LINKED ISSUE #987: Micropython autocomplete problems
When using the micropython backend, the autocomplete feature of the interpreter fails when the autocompleted object doesn't exist in the current context. For example:

`>>> asdf.<TAB>`

Thonny throws many pop-ups (on macOS sometimes even as many as 5, all with the same content; on Windows 7 only the first one) with errors:
* `AttributeError: 'InlineResponse' object has no attribute 'source'`
* `TypeError: 'NoneType' object is not iterable`.

This doesn't happen when the standard backend is set.

adzierzanowski:


aivarannamaa:
Thank you for the PR and sorry for the late reply! I need to take time and pay attention to this topic, but I'm busy with other things at the moment. I hope to find time for this next week.

aivarannamaa:
Looks good, thanks!

Issue #1041, opened 2019-12-28
save all button


kr-g:
save all button for saving all changed files, or in combination with F5 / Run, would be nice

aivarannamaa:
I'd prefer not adding another button, but saving all files on F5 looks like a good idea!

hatlesshacker:
I can do that. Mind if I submit a pull request?

aivarannamaa:
@hatlesshacker, you are welcome!

hatlesshacker:
see #1053, not #1052.
I had previously set Shift-Q as the Save-all keyboard shortcut, which closes the window on some systems.

PR #1052, opened 2020-01-08
Added 'Save All' feature

hatlesshacker:
https://github.com/thonny/thonny/issues/1041

### DIFF
--- a/thonny/code.py
+++ b/thonny/code.py
@@ -561,6 +561,19 @@ class EditorNotebook(ui_utils.ClosableNotebook):
         )
 
         get_workbench().add_command(
+            "save_all_files",
+            "file",
+            _("Save All files"),
+            self._cmd_save_all_files,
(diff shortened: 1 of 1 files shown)


thonny/code.py
```diff
@@ -560,6 +560,19 @@ def _init_commands(self):
             include_in_toolbar=True,
         )
 
+        get_workbench().add_command(
+            "save_all_files",
+            "file",
+            _("Save All files"),
+            self._cmd_save_all_files,
+            caption=_("Save All files"),
+            default_sequence=select_sequence("<Control-q>", "<Command-q>"),
```
aroberge: Is this correct?  Control-q is the default on Windows for closing a program - not for saving all files.

hatlesshacker: Exactly. I forgot about that. See #1053

Issue #1055, opened 2020-01-10
Implement ESP Flash Mode  (--flash_mode )


rlangoy:
It would be very nice if  the flash dialogbox for esp8266, implements an options for [flash_mode](https://github.com/espressif/esptool#flash-mode---flash_mode--fm)

The WeMos D1 mini  will only boot when flashed with --flash_mode dio, and therefor it can't be flashed directly using Thonny  and the  micropython.org firmware for esp8266.

There are other esp8266 modules, including the ESP-12E modules and some (not all) NodeMCU boards, are dual I/O.

PR #1056, opened 2020-01-11
Implement ESP Flash Mode (--flash_mode ) #1055

rlangoy:
Issue  #1055
Implements an options for flash_mode in the ESP flash dialogbox  

[Proposed ESP Flash dialog](https://raw.githubusercontent.com/rlangoy/thonny/master/docs/InstallFirmwareDlg.PNG)

Reason:
The WeMos D1 mini will only boot when flashed with --flash_mode dio, and therefor it can't be flashed directly using Thonny and the micropython.org firmware for esp8266.

There are other esp8266 modules, including the ESP-12E modules and some (not all) NodeMCU boards, are dual I/O.

InstallFirmwareDlg


aivarannamaa:
Lovely! Thank you very much!

Issue #1021, opened 2019-12-09
Warn against running from a library directory


aivarannamaa:
See #1017

hatlesshacker:
By "Library directories" are you referring to only the directory where the python packages are installed, or are there additional locations where Thonny should not be run?

aivarannamaa:
I  mostly meant `site-packages` and `dist-packages` directories.

hatlesshacker:
Ok, I'll try to submit a pull request

aivarannamaa:
I guess the nice location for this warning would be "Assistant" view. If you agree, then see `thonny.plugins.misc_analyzers` for an example. I suppose the new warning can also go into same module.

hatlesshacker:
See #1063 @aivarannamaa

aivarannamaa:
This is actually non-issue.

Issue #1017, opened 2019-12-07
Backend Error on startup


thamilton533:
I have been using a Thonny for a few weeks now and last night I started getting the error below.

This error occurs when the program starts, not when I run script.

```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/thonny/backend_launcher.py", line 76, in <module>
    VM().mainloop()
  File "/usr/lib/python3/dist-packages/thonny/backend.py", line 122, in __init__
    self._load_shared_modules()
  File "/usr/lib/python3/dist-packages/thonny/backend.py", line 298, in _load_shared_modules
    self.load_modules_with_frontend_path(["parso", "jedi", "thonnycontrib", "six", "asttokens"])
  File "/usr/lib/python3/dist-packages/thonny/backend.py", line 308, in load_modules_with_frontend_path
    import_module(name)
  File "/usr/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/usr/lib/python3/dist-packages/jedi/__init__.py", line 41, in <module>
    from jedi.api import Script, Interpreter, set_debug_function, \
  File "/usr/lib/python3/dist-packages/jedi/api/__init__.py", line 29, in <module>
    from jedi.api.project import get_default_project
  File "/usr/lib/python3/dist-packages/jedi/api/project.py", line 2, in <module>
    import json
  File "/usr/lib/python3/dist-packages/flask/json/__init__.py", line 13, in <module>
    from flask.globals import current_app, request
  File "/usr/lib/python3/dist-packages/flask/__init__.py", line 19, in <module>
    from jinja2 import Markup, escape
  File "/usr/lib/python3/dist-packages/jinja2/__init__.py", line 33, in <module>
    from jinja2.environment import Environment, Template
  File "/usr/lib/python3/dist-packages/jinja2/environment.py", line 15, in <module>
    from jinja2 import nodes
[...]

PR #1063, opened 2020-01-17
Warning on running from a library directory

hatlesshacker:


aivarannamaa:
The problem in #1017, which motivated #1021 was actually subtly different -- in this case the *backend* process was started in library directory, probably because the user navigated into this directory via Thonny's file browser.

Your PR would catch the case where Thonny's *frontend* process (the GUI) is started from a library directory. 

In retrospect I see that #1021 is simply misleading. I focused on wrong thing while analyzing #1017. The problem was not that the working directory was under a library directory -- the problem was that the directory contained a file, which shadowed an important library module. This can happen anywhere.

I appreciate your effort on this, but I feel that such check is not actually useful. So, I'm rejecting this PR (unless you have something more to add). Sorry!

hatlesshacker:
I understand; No problem 👍

PR #1086, opened 2020-02-13
Improve English in About screen

phr0gger:


aivarannamaa:
Thanks! :)

aivarannamaa:
Oops, I forgot that this will break all existing translations. I'll revert your commit and include your suggestion as English translation instead of the base string in the source code.

PR #1102, opened 2020-03-02
The translattion of "Interpreter" for zh_TW should be  "直譯器"

codemee:
The translation of "interpreter" in current version is "直譯器" or "直釋器", it should be unified to "直譯器". That's what we call "interpreter" here in Taiwan--the place where Tradition Chinese is mainly used.

### DIFF
--- a/thonny/locale/zh_TW/LC_MESSAGES/thonny.po
+++ b/thonny/locale/zh_TW/LC_MESSAGES/thonny.po
@@ -127,11 +127,11 @@ msgstr "自動完成"
 
 #: thonny/plugins/backend_config_page.py:43
 msgid "Which interpreter or device should Thonny use for running your code?"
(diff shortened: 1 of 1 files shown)


aivarannamaa:
Did you also update the translation in Poeditor (https://poeditor.com/join/project/Gh188fdYH6) ?

codemee:
I've updated the translation in POEditor.Do I need to update the .po and .mo files?

aivarannamaa:
> I've updated the translation in POEditor.Do I need to update the .po and .mo files?

No, this is not required. I'll download all modified languages before each release anyway.

aivarannamaa:
Thank you for your contribution!
Hopefully other users of your language agree with your term :)

codemee:
OK, thanks.

PR #1107, opened 2020-03-05
Convert keypad movement events to equivalent non-keypad ones.

eliotb:
Improves experience for users that only have numpad/keypad movement keys, lacking standalone versions of same.

### DIFF
--- a/thonny/tktextext.py
+++ b/thonny/tktextext.py
@@ -195,6 +195,7 @@ class EnhancedText(TweakableText):
         self._last_event_kind = None
         self._last_key_time = None
 
+        self._bind_keypad()
         self._bind_editing_aids()
         self._bind_movement_aids()
         self._bind_selection_aids()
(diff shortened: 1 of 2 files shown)


thonny/tktextext.py
```diff
@@ -249,6 +250,23 @@ def dispatch(event):
...
+            '<KP_Home>': '<Home>',
+            '<KP_End>': '<End>',
+            '<KP_Next>': '<Next>',
+            '<KP_Prior>': '<Prior>',
+            '<KP_Enter>': '<Return>',
+        }
+        for i in kmap:
+            def mfunc(event, key=i):
+                self.event_generate(kmap[key], **{'state': event.state})
+            self.bind(i, mfunc)
```
aivarannamaa: In Windows it gives me `_tkinter.TclError: bad event type or keysym "KP_Left"`. Numpad left keycode is just `Left` in Windows. You should catch and silence the binding errors.

I'm also worried that on some platforms this approach may generate double movement actions (one for KP_Left and another for generated Left). The standard approach is ending the event handler with `return "break"` -- this should stop the event propagation.

aivarannamaa: Also, please add link to the issue as comment inside `bind_keypad`!

aivarannamaa:
Thank you for the contribution! Please see my comments in the code!

aivarannamaa:
BTW. Do you have IDLE installed? Does your numpad movement keys work there?

Which Tk version do you have? ("Thonny => Help => About" and "IDLE => Help => About IDLE")

eliotb:
> BTW. Do you have IDLE installed? Does your numpad movement keys work there?

Just installed, and no, numpad movement doesn't work there either.

> Which Tk version do you have? ("Thonny => Help => About" and "IDLE => Help => About IDLE")

Python 3.7.3, Tk 8.6.9

thonny/tktextext.py
eliotb: I created a new pull request rebased and squashed onto current master branch.

aivarannamaa:
Thanks! I'll test it in my Ubuntu at work.

PR #1187, opened 2020-05-05
Force to de-assert DTR# and RTS# at Thonny startup

webduino-cn:
Some Arduino-style boards that connect DTR# or RTS# to RESET directly.  And sometimes, the driver of the serial port asserts the DTR# and RTS# (Verifyied with CH340G and CP2102). So I suggest that we de-assert DTR# and RTS# at the plugin starting.

### DIFF
--- a/thonny/plugins/micropython/serial_connection.py
+++ b/thonny/plugins/micropython/serial_connection.py
@@ -15,6 +15,8 @@ class SerialConnection(MicroPythonConnection):
 
         try:
             self._serial = serial.Serial(port, baudrate=baudrate, timeout=None)
+            self._serial.dtr = 0
+            self._serial.rts = 0
         except SerialException as error:
             err_str = str(error)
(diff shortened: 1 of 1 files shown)


aivarannamaa:
I'm not familiar with this topic. Can you please explain more, what means de-assert?

Which problem would this change solve?

The constructor of the `Serial` class seems to disable `rtscts` and `dsrdtr` by default (https://pythonhosted.org/pyserial/pyserial_api.html#serial.Serial). How does this relate to your change?

aivarannamaa:
@webduino-cn , are you still interested in this PR? If you are, then please answer my question above!

aivarannamaa:
Is it related to the comment given here: https://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.Serial.open ?

aivarannamaa:
I merged as it didn't seem to do any harm and hopefully it is useful in some cases.

PR #1203, opened 2020-05-23
Bash install - do everything inside a new directory

lurch:
This prevents deleting any `thonny` directory that already existed!


aivarannamaa:
Thanks!

PR #1209, opened 2020-05-24
Start file-open-dialog in same dir as current file

### LINKED ISSUE #1208: Feature request: file-open dialog enhancement
It would be nice if when I choose "Open..." that Thonny defaulted the file-selection dialog to be in the same directory as the file I'm currently editing. It would make it much easier to open several files for the same project!

lurch:
Fixes #1208

Well that was easier than expected :slightly_smiling_face:

### DIFF
--- a/thonny/code.py
+++ b/thonny/code.py
@@ -695,9 +695,13 @@ class EditorNotebook(ui_utils.ClosableNotebook):
             return None
 
         if node == "local":
-            path = askopenfilename(
-                filetypes=_dialog_filetypes, initialdir=get_workbench().get_local_cwd()
-            )
+            initialdir = get_workbench().get_local_cwd()
+            if (
+                self.get_current_editor() is not None
+                and self.get_current_editor().get_filename() is not None
+            ):
+                initialdir = os.path.dirname(self.get_current_editor().get_filename())
+            path = askopenfilename(filetypes=_dialog_filetypes, initialdir=initialdir)
         else:
             assert node == "remote"
             target_path = ask_backend_path(self.winfo_toplevel(), "open")


aivarannamaa:
I think there are situations where current dir would be better choice than the directory of current file -- eg. when the user starts closing tabs from current directory (with the intent of opening another from this directory) and ends up with current file being in another directory.

At the same time, there are definitely cases where your approach would be more helpful (eg. when user is just browsing some files without executing them ie. the current directory is not changing).

What do you think of a history-based approach, where each Open, Save As and `%cd` action (invoked by running a script or entering a directory in the file browser) would update a configuration option (`file.last_interesting_dir`)?

aivarannamaa:
For sure, your approach is easiest to implement and easiest to explain...

aivarannamaa:
BTW. Did you notice the current directory being changed by entering a directory in the file browser or by running a script from another directory?

Hermann12:
I work on different projects with thonny. I wish I can have a menu like "projects", where I can preset different folders for each project. So I can easy access the right folder, when I come back day's later.
If I step into one folder thonny remember, which files was last time open (maybe last 3). Suits maybe to this discussion here?

aivarannamaa:
@Hermann12, this would be nice to have, but Thonny's target group (the beginners) usually have single-file "projects", so I adding a support for separate project layer would complicate matters without bringing any benefits to most of the users. Besides, there are many more pressing issues in the TODO list.

I think those users, who work with several multi-file projects, would be more happy with a more advanced IDE (eg. PyCharm or PyDev).

lurch:
> What do you think of a history-based approach, where each Open, Save As

Well, if you Open a file, and still have that file in the current tab, then my approach solves this too. Same for Save As :wink: 

I think the problem with the `file.last_interesting_dir` approach is that it's not obvious what `file.last_interesting_dir` is currently set to, or how a user would change it?

aivarannamaa:
> I think the problem with the file.last_interesting_dir approach is that it's not obvious what file.last_interesting_dir is currently set to, or how a user would change it? (i.e. it might feel "mysterious" to a user what does and doesn't cause it to change)

You're right. I'll merge your PR. Thank you!

Issue #1079, opened 2020-01-28
Venv interpreter soft link gets resolved in macOS


aivarannamaa:

kalhauge:
I second this :).  Alternatively, It would be nice with a choose virtual environment button instead of only being able to create a new one.

aivarannamaa:
Thanks for your suggestion! Still, I'd prefer to keep virtual env selection as it is now -- by selecting its interpreter (not its root directory). Unfortunately macOS file chooser always resolves symbolic links, so it doesn't allow selecting venv interpreter. 

I added a warning about this and made it possible to bypass the file selection dialog and enter the interpreter path directly. This inconvenience appears only once per externally created venv -- next time you can already choose it from the list.

I also took the opportunity to refactor the UI of the configuration page.

aivarannamaa:
Please reopen the issue, if you don't agree with my solution!

kalhauge:
Works really well! Great job :)

PR #1358, opened 2020-10-06
Partial solution to #1079

kalhauge:
While this still does not fix the problem listed in #1079. It is an alternative solution, with an easier approach to using 
existing virtual environments.

Tested on MacOS but still requires some error handling.


kalhauge:
Made obsolete by c4fdd0a

PR #1381, opened 2020-10-21
Add default black fg color to tooltips for macOS dark mode

adzierzanowski:
On macOS in the dark mode, the tooltips are white text on a yellowish background, rendering them unreadable. This adds the default foreground color set to black.


aivarannamaa:
Thank you!

PR #1382, opened 2020-10-21
Hide underscored names from autocomplete suggestions unless user already typed '_'

adzierzanowski:
@aivarannamaa 

Hi, I would like to discuss a feature that would allow names that start with underscore character not to be on the autocompletion list.

I've added the option "Allow names starting with '_' char in completions" which is set to `True` by default. The reasoning behind this option is that in Python, private names begin with this character. Those private names add clutter especially for beginners who wouldn't need to use them (at least in the project I'm working on now).

As for now, this is just a draft. I'd like to be able to force the completion if at least one underscore has already been typed by the user. Another option for solving the issue is just to move the underscored completions to the end of the list. I'd like to know your opinion on that.

Regards,
Aleksander

### DIFF
--- a/thonny/plugins/autocomplete.py
+++ b/thonny/plugins/autocomplete.py
@@ -71,10 +71,10 @@ class Completer(tk.Listbox):
             self._present_completions(msg.completions)
 
     def _present_completions(self, completions_):
-        if get_workbench().get_option('edit.tab_complete_show_private'):
-          completions = completions_
+        if get_workbench().get_option("edit.tab_complete_show_private"):
+            completions = completions_
         else:
-          completions = [c for c in completions_ if not c.get('name', '_').startswith('_')]
+            completions = [c for c in completions_ if not c.get("name", "_").startswith("_")]
 
         self.completions = completions
 
--- a/thonny/plugins/autocomplete.py
+++ b/thonny/plugins/autocomplete.py
@@ -77,10 +77,10 @@ class Completer(tk.Listbox):
         # if it doesn't - don't show names starting with '_'
         source = self.text.get("1.0", "end-1c")
         try:
-            last_source_chunk = re.split(r'\.|\s', source)[-1]
+            last_source_chunk = re.split(r"\.|\s", source)[-1]
         except IndexError:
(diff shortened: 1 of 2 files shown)


aivarannamaa:
I like the idea that you need to type the underscore in order to show private names.

But I would omit the option from the configuration UI. Adding single option is not a problem by itself, but if we are too generous, the configuration page may come too crowded.

adzierzanowski:
OK, I changed it so the underscored names appear only when user typed an underscore. I also removed the option.

I took last part of current line delimited with a dot or whitespace but Python has some more delimiters for its grammar.

https://docs.python.org/3/reference/lexical_analysis.html#delimiters

Do you think it's a good idea to include them as well?

thonny/plugins/autocomplete.py
```diff
@@ -70,7 +72,21 @@ def _handle_backend_response(self, msg):
         else:
             self._present_completions(msg.completions)
 
-    def _present_completions(self, completions):
+    def _present_completions(self, completions_):
+        # Check if autocompeted name starts with an underscore,
+        # if it doesn't - don't show names starting with '_'
+        source = self.text.get("1.0", "end-1c")
```
aivarannamaa, line 78: It looks like you are assuming, the user is appending to the source, but in reality, the insertion point can be in the middle of the source.

I suggest you ask for text from the start of the current line up to the insertion cursor: `self.text.get("insert linestart", "insert")`

aivarannamaa (changes requested):
please check my review comment about fetching relevant piece of source.

thonny/plugins/autocomplete.py
```diff
@@ -70,7 +72,21 @@ def _handle_backend_response(self, msg):
...
+        try:
+            last_source_chunk = re.split(r"\.|\s", source)[-1]
```
aivarannamaa, line 80: What about splitting on the non-word characters (`\W`)?

adzierzanowski:
Both suggestions are of course very valid. I dare to think the feature works now as expected.

aivarannamaa:
Looks good. Thank you!

PR #1415, opened 2020-11-14
Ctrl+m shortcut to open recent files.

OHHH-YEHH:


aivarannamaa:
Thank you for the PR!

Do I understand correctly, that you want it for opening the files from the previous session?

Did you know about setting "Reopen all files from previous session" under "Tools => Options => General"? Wouldn't this solve your case?

OHHH-YEHH:
Initially, I wanted to build the functionality provided by Ctrl+Shift+t that we see on chrome i.e simply put suppose there are 4 tabs open in the editor, and accidentally / on the purpose you close a particular tab but later found that you want to open that particular file/tab. To do so,  you have to first go to the File menu and then visit the recent submenu to open that particular file.

With now Ctrl+m shortcut, we will be able to open the first file from the recent files submenu.

aivarannamaa:
I see.

I appreciate your effort of digging into Thonny's code, but for me this functionality doesn't look important enough to warrant a new menu item and a shortcut. I must keep in mind that this is a beginner's IDE and I need to be conservative when accepting new features. Otherwise the menus can get too crowded and I could be out of good shortcuts if I need one.

On the other hand, I encourage you to publish this enhancement (maybe bundled with something else) as a plug-in. The general overview of writing plug-ins for Thonny is https://github.com/thonny/thonny/wiki/Plugins.

Here you can see an example of adding a menu item from a plug-in: https://github.com/thonny/thonny/blob/master/thonny/plugins/commenting_indenting.py (this is a built-in plug-in, but the approach is the same with 3rd party plug-ins). Don't hesitate to ask for help if you get stuck!

PR #1473, opened 2020-12-08
Missing terms in POEditor

codemee:
It seems there're some terms not found in POEditor as shown below

圖片

I'd like to translate these terms into Traditional Chinese.


aivarannamaa:
Thank you for the suggestion! I created #1484 for missing terms. 

I'll pull all translations from POEditor before next release, so no PR is required anymore. I now changed the document accordingly.

PR #1647, opened 2021-02-23
Fixing the print margin for numbers

### LINKED ISSUE #1646: line number field not wide enough when printing .py file with 1000 or more lines
When printing a Python program listing with 1000 or more lines, the line number field "spills over" the least-significant digit of the line number to the next line, which effectively double-spaces subsequent lines of the program. For example, a listing will look like this:
...
998  # lines 999 and below are fine
999  # but after this, things mess up
100
    0  # line 1000 goes here
100
    1  # line 1001 is here

I've tried changing Thonny's margins, but that has no effect on the line numbers. 

I could not find this addressed anywhere, but could someone help me with a solution or a work-around, please?  Thanks!!

gagzilla:
The earlier version had a fixed margin of 2em. Switching to a minimum of 2em- this should scale to accommodate longer text for numbers beyond 1000. Fixes #1646

### DIFF
--- a/thonny/plugins/printing/template.html
+++ b/thonny/plugins/printing/template.html
@@ -33,7 +33,7 @@ code:before {
     border-right: solid gray 1px;
     display: inline-block;
     text-align: right;
-    min-width: 2em;
+    min-width: 3em;
     font-style: italic;
     -webkit-user-select: none;
 }
--- a/thonny/plugins/printing/template.html
+++ b/thonny/plugins/printing/template.html
@@ -33,7 +33,7 @@ code:before {
     border-right: solid gray 1px;
     display: inline-block;
     text-align: right;
-    width: 2em;
+    min-width: 2em;
     font-style: italic;
     -webkit-user-select: none;
 }


aivarannamaa:
Thank you! That was quick! :)

I noticed that at least in Chrome, there is still a small glitch between line 999 and 1000:

image

My CSS is rusty. Do you have ideas how to fix this as well?

aivarannamaa:
Just saying, that I'm happy to merge the PR as it is, if you don't have a good solution for this minor glitch, because it is already so much better.

gagzilla:
To align the right margin at 999 and 1000 we'd have to fix the width to be consistent (which is what the original problem was) or alternatively use something like a table for each line (which is a bit of an overkill) By using 2em - we see this problem at 1000 mark, but if we bump that min-width to say 3 or 4 em, we can push this out for a much higher number of lines (10000, or even 100000) which hopefully isn't that typical. The only side-effect would then be that the shorter files will have a slightly wider numbers margin.

aivarannamaa:
Thank you for the explanation!

PR #1709, opened 2021-03-16
Hide extra dependencies in plugin manager

mristin:
This patch hides the dependencies of a plug-in which refer to
`extra == *`.

These dependencies can be confusing for the user as they are actually
not installed. For example, a plug-in might use `black` to format the
code and mark it as `dev` dependency, but Thonny will not install it
when acquiring the plug-in.

We added `packaging` to requirements instead of importing
`packaging.markers` from `pkg_resources` or `pip` (as library). This is
much easier to trace instead of using `__import__` builtin.

Fixes #1709.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,4 +6,3 @@ docutils>=0.14
 mypy>=0.670
 asttokens>=1.1
 Send2Trash>=1.4
-packaging>=20.9
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ docutils>=0.14
 mypy>=0.670
 asttokens>=1.1
 Send2Trash>=1.4
+packaging>=20.9
--- a/thonny/plugins/pip_gui.py
+++ b/thonny/plugins/pip_gui.py
@@ -12,6 +12,9 @@ from tkinter import messagebox, ttk
 from tkinter.messagebox import showerror
 from typing import List, Union, Dict, Tuple
 
+import pyparsing
+import packaging.markers
+
 import thonny
 from thonny import get_runner, get_workbench, running, tktextext, ui_utils
 from thonny.common import InlineCommand, is_same_path, normpath_with_actual_case, path_startswith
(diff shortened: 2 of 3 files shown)


thonny/plugins/pip_gui.py
```diff
@@ -523,7 +526,49 @@ def write_att(caption, value, value_tag=None):
         if info.get("requires_dist", None):
             # Available only when release is created by a binary wheel
             # https://github.com/pypa/pypi-legacy/issues/622#issuecomment-305829257
-            write_att(tr("Requires"), ", ".join(info["requires_dist"]))
+            requires_dist = info["requires_dist"]
+            assert isinstance(requires_dist, list)
```
mristin, line 530: @aivarannamaa please let me know if you do not want to enforce this kind of type safety at runtime.

mristin:
@aivarannamaa the pull request is ready for review.

requirements.txt
```diff
@@ -6,3 +6,4 @@ docutils>=0.14
 mypy>=0.670
 asttokens>=1.1
 Send2Trash>=1.4
+packaging>=20.9
```
aivarannamaa, line 9: This won't do, as Debian Buster only has 19.0

mristin: TLet me see if 19.0 also works. I've just pinned the latest, admittedly a bit too sloppy.

aivarannamaa:
Thank you!

I'm concerned about introducing a new dependency, though. If I only distributed Thonny as binary bundle, it wouldn't matter, but Thonny is also published in several Linux repos and its deps must be resolvable there and it should work with whatever version the distros offer. So, for this improvement I'd rather not introduce a new dependency. 

Would it be too hard to parse the specs by hand?

mristin:
> Would it be too hard to parse the specs by hand?

Let me have a look at the code of packaging.markers.

mristin:
@aivarannamaa I removed the `packaging` dependency with a regular expression and added a unit test. Please let me know if you have a different style on regular expressions. This might not catch some of the edge cases in the wild, but should work in the majority of the cases.

thonny/plugins/pip_gui.py
aivarannamaa: It's OK.

aivarannamaa (approved):
Now I'm happy. Thank you! :)

PR #1758, opened 2021-04-11
[TRANS] Help Content translated to korean. Index files and titles each pages.

fabichoi:
Translated index page and titles each pages.
Descriptions of each pages will be translated until May, 2021.

### DIFF
--- a/thonny/locale/ko_KR/HELP_CONTENT/birdseye.rst
+++ b/thonny/locale/ko_KR/HELP_CONTENT/birdseye.rst
@@ -1,24 +1,24 @@
 버드아이에 대해
 ==========================
 
-Birdseye is a cool Python debugger by Alex Hall, which records the values of expressions 
-when the program runs and lets you explore them after the program completes. See
-`https://birdseye.readthedocs.io <https://birdseye.readthedocs.io>`_ for more info.
+Birdseye는 Alex Hall이 만든 멋진 Python 디버거로, 프로그램이 실행될 때 표현식의 값들을 기록하고
+프로그램이 완료된 후 이 값들을 탐색 할 수 있습니다. 더 많은 정보는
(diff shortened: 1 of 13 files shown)


fabichoi:
@aivarannamaa is it wrong way to add translating help contents? let me know how to do merge master branch if you are okay.

aivarannamaa:
Sorry for late reply!

I'm curious, why do you want to translate the titles without translating the main text? 

The downside of this partial translation is that whenever I update the main English text under thonny/plugins/help, the Korean users don't see it until you update the Korean copies. (Of course, it would be the same if you translated everything, but in that case the benefit of full Korean text might outweigh this.)

aivarannamaa:
What about translating the index page only? In this case the user would see the latest content of the topic pages.

fabichoi:
Oh.. I see what you meant. You think I should translate for all of help contents.
I'll translate all of help contents until end of this month. (April, 2021)
And then, leave a comment for PR again. 😃

aivarannamaa:
> You think I should translate for all of help contents.

Either this, or translate only the index page.

aivarannamaa:
> > You think I should translate for all of help contents.

> Either this, or translate only the index page.

... or only selected pages. My main idea was that once you create a new rst file under a locale, the users of this language stop automatically receiving updates made to the English rst.

fabichoi:
@aivarannamaa I'm done with translating Korean for all of docs files. 
Can you check my tasks when you are not busy, please? 😄

aivarannamaa:
Thanks! Unfortunately I don't know Korean so I can't grade your work, but I believe it's good :)

fabichoi:
Thanks a lot! lol. I'll keep in touch this project and update translation to Korean :D

Issue #1751, opened 2021-04-04
Task/Todo textboxes based on code?


mristin:
Hi @aivarannamaa,
Please apologize if this feature has been already requested or implemented. I couldn't find a relevant issue.

We are looking into how to use contracts for educational purposes (see [thonny-icontract-hypothesis] and [thonny-crosshair]). Consider a classroom exercise. I thought that it would be practical if the *teacher* could write the interfaces of the functions (including the contracts) and mark the blank areas in the code (*e.g.*, a "solution area"), while the *pupils* would fill out the blanks. To avoid confusion, A thonny plug-in would prevent editing of the non-blank areas.

Do you know if somebody already looked into it? If not, how hard would it be to implement such a plug-in in your opinion? (Would you be perhaps interested to help me implement it if nobody looked into it thus far? I can do the "hard lifting", but I'd need some guidance.)

[thonny-icontract-hypothesis]: https://github.com/mristin/thonny-icontract-hypothesis/
[thonny-crosshair]: https://github.com/mristin/thonny-crosshair/

aivarannamaa:
I'm not aware of any attempts to do something similar in Thonny's context, but it shouldn't be too hard to do.

Thonny's editor's superclass has couple of methods which can be helpful here, see `thonny.shell.BaseShellText.intercept_insert` and `intercept_delete` for an example. You should be able to use Text's tags (https://tkdocs.com/shipman/text-tag.html) for bringing out the editable parts visually. 

Unortunately, there is no clean way for getting the modified behavior into Thonny's editor class. The best way I can think of, is simply patching the class. See the last 2 lines in `thonny.plugins.autocomplete` for an example.

I'm happy to help when you get stuck.

Out of curiousity, why can't you just tell the students to edit only the marked areas in the code? What kind of confusion do you expect?

mristin:
> Unortunately, there is no clean way for getting the modified behavior into Thonny's editor class. The best way I can think of, is simply patching the class. See the last 2 lines in thonny.plugins.autocomplete for an example.

Thanks, I'll see how that works!

> Out of curiousity, why can't you just tell the students to edit only the marked areas in the code? What kind of confusion do you expect?

Think high school students in puberty ;-). This is not an issue for the university students, but it makes the life easier for the high school students that they are sure they can not "destroy" the scaffolding code and clearly see what needs to be filled out.

I'll keep you posted how it goes.

mristin:
Hi @aivarannamaa,
[...]

PR #1878, opened 2021-06-25
Sign script

nicolasdespres:
Hi,

First of all, thank you very much for developing Thonny. I have been using it for introducing Python to high-school students and it has worked seamlessly. Congrats!

One of the main issue my colleague and I have faced this year is ensuring students have done their homework... This problem is not related to Thonny but is general to all digital assignments with young students.

The solution I have found is to include a finger print based on the edit history (insertion and deletion) into each file. Such a finger print could be used to check slightly modified copied files, copy-paste, distinguish two students ending up with same content, etc...

I have tried to make it the least intrusive possible with the basic knowledge I have of Thonny's internal.

Best regards,

### DIFF
--- a/thonny/codeview.py
+++ b/thonny/codeview.py
@@ -12,6 +12,7 @@ from thonny import get_workbench, roughparse, tktextext, ui_utils
 from thonny.common import TextRange
 from thonny.tktextext import EnhancedText
 from thonny.ui_utils import EnhancedTextWithLogging, scrollbar_style, ask_string
+from thonny.text_edit_finger_print import insert_finger_print, extract_finger_print, FingerPrintError
 
 _syntax_options = {}  # type: Dict[str, Union[str, int]]
 # BREAKPOINT_SYMBOL = "•" # Bullet
@@ -206,8 +207,14 @@ class CodeView(tktextext.EnhancedTextFrame):
     def set_file_type(self, file_type):
         self.text.set_file_type(file_type)
 
-    def get_content_as_bytes(self):
+    def get_content_as_bytes(self, with_finger_print=False):
         content = self.get_content()
+
+        if with_finger_print:
+            content = insert_finger_print(
+                content,
+                self.text._edit_history.history,
+                newline=self._original_newlines)
 
         # convert all linebreaks to original format
         content = OLD_MAC_LINEBREAK.sub(self._original_newlines, content)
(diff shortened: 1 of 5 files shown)


aivarannamaa:
Thanks! Sounds interesting, but currently I find it a bit to follow your changes.

Could you please give me an short overview, about what will happen when the user edits or pastes code and how this helps detecting cheating?

For example, what happens, when the one student copies another student's work and makes some cosmetic changes to it? Can you detect cheating in this case?

I know that some teachers require students to turn in the action logs (https://github.com/thonny/thonny/wiki/User-action-logs). How would this compare to your approach?

nicolasdespres:
Thanks for your reply !

I was not aware of the User-action-logs. The main difference I see with my approach is that the user action log is attached to a session whereas the finger print is related to a single file. In addition, the finger print is included in the python file, so there is no extra file to hand-in. Action log contains timestamp which I should probably add to the finger print.

The finger print is the edit action history, serialized in json, compressed and encoded in base64. It is then inserted as a comment at the end of the python script.

The idea is to collect the python files from all students and to run a "cheat" analysis tool on this collection (potentially including files from previous years too).

Copy-paste will appear as one big insert action.

Cosmetic change will most likely appear only as space/tabs/newline insertion.

Variable renames could be detected by checking if only characters belonging to variable names are deleted/inserted.

I have not developed such analysis tool yet. It will certainly need to be tweaked and checked against real-world data.

aivarannamaa:
Have you considered publishing your work as Thonny's plug-in (https://github.com/thonny/thonny/wiki/Plugins)? It would give you more freedom in releasing the updates. You could take a look at https://github.com/thonny/thonny/issues/1751 for an example. It may be more inconvenient in places, but we could agree on introducing some new extension points, if required.
 
I don't rule out the possibility to include your work in Thonny's main code, but I don't want to do this before it feels useful and desirable for other teachers as well. In particular, the analysis tool should be publicly available.

nicolasdespres:
Thanks for the tips. I will implement it as a plugin, and I will publish the analysis tool when it is ready. 

After some tests, I think I would need three extensions:
- a `EditorTextDestroyed` event as the opposite of the `EditorTextCreated` event. Currently only a sequence "<Button-1>" on the `EditorNotebook` is emitted.
- a save and load hooks to insert/extract the finger print

Without the `EditorTextDestroyed`, there is no clean way to detect when a tab is closed, and the plugin will leak some objects.

Without the save hook, the file will be written twice after a save (re-detecting the encoding and not taking advantage of the newline cleanup) and it won't work with remote file.

Without the load hook, the content of the `CodeViewText` widget will be reset twice.

Do you think these extensions are reasonable ?

I could try to implement them as a separate pull-request.

aivarannamaa:
> Do you think these extensions are reasonable ?

Yes.

> I could try to implement them as a separate pull-request.

Please do! I'll wait with releasing version 3.3.12 until this.

nicolasdespres:
Closed in favor to #1891

PR #1884, opened 2021-07-03
recognize method and function calls for syntax coloring

campbelljc:
While working on my own theme for Thonny (porting a theme from jEdit), I noticed that there is no recognition for method or function calls as is the case in other editors. So I have added a regular expression to catch these. As to speed, I am not sure, but anecdotally I tried editing a large (5000+ lines) file and did not notice any issue.  With the regular expressions, I did not include optional spaces before opening parenthesis, I am not sure if that would be needed.

Screen Shot 2021-07-03 at 3 06 43 PM

### DIFF
--- a/thonny/plugins/coloring.py
+++ b/thonny/plugins/coloring.py
@@ -99,6 +99,8 @@ class SyntaxColorer:
             "definition",
             "function_call",
             "method_call",
+            "class_definition",
+            "function_definition",
         }
         self.multiline_tags = {"string3", "open_string3"}
         self._raise_tags()
@@ -112,7 +114,7 @@ class SyntaxColorer:
         self.text.tag_raise("open_string3")
         self.text.tag_raise("open_string")
         self.text.tag_raise("sel")
-        self.text.tag_raise("builtin")
+        self.text.tag_raise("builtin", "function_call")
         """
         tags = self.text.tag_names()
         # take into account that without themes some tags may be undefined
@@ -190,6 +192,15 @@ class SyntaxColorer:
                                 id_match_start, id_match_end = id_match.span(1)
                                 self.text.tag_add(
                                     "definition",
+                                    start + "+%dc" % id_match_start,
+                                    start + "+%dc" % id_match_end,
+                                )
+                                if token_text == "def":
+                                    tag_type = "function_definition"
+                                else:
+                                    tag_type = "class_definition"
+                                self.text.tag_add(
+                                    tag_type,
                                     start + "+%dc" % id_match_start,
                                     start + "+%dc" % id_match_end,
                                 )
--- a/thonny/plugins/coloring.py
+++ b/thonny/plugins/coloring.py
@@ -50,6 +50,8 @@ class SyntaxColorer:
             STRING_CLOSED,
             STRING_OPEN,
             TAB,
+            FUNCTIONS,
+            METHODS,
         )
 
         self.uniline_regex = re.compile(
@@ -69,7 +71,11 @@ class SyntaxColorer:
             + "|"
             + STRING_OPEN
             + "|"
-            + TAB,
+            + TAB
+            + "|"
+            + FUNCTIONS
+            + "|"
+            + METHODS,
             re.S,  # @UndefinedVariable
         )
 
@@ -91,6 +97,8 @@ class SyntaxColorer:
             "number",
             "builtin",
             "definition",
+            "functions",
+            "methods",
         }
         self.multiline_tags = {"string3", "open_string3"}
         self._raise_tags()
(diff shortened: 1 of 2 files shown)


aivarannamaa:
Thank you for the suggestion!

Is it common to have different coloring for function calls? I've seen different coloring of attributes (this includes methods in method calls), but not calls per se.

Do you think it's worth including separate coloring classes for functions and methods?

Thonny also gives separate coloring class for built-in names (eg `len`). When they are called, it would get two classes -- "builtin" and "function". Which should take precedence, when both are assigned a different foreground color?

You only considered latin letters for function names. In non-English countries, students are likely to use other letters in names as well. I suggest using `\w` here. I think it's ok to allow the name to start with number here, because number literals can't be followed by an opening paren.

You named the classes in plural. Other classes are in singular, so I suggest to use singular here as well.

campbelljc:
> Thank you for the suggestion!
> 
> Is it common to have different coloring for function calls? I've seen different coloring of attributes (this includes methods in method calls), but not calls per se.
> 
> Do you think it's worth including separate coloring classes for functions and methods?
>

I'm not sure how prevalent it is to have coloring for function calls. I know that the Highlight tool (http://www.andre-simon.de), from which I was porting a theme, does have support for it.

As to whether it should be separated from methods, the theme I have ported has the same color for both, but I separated them as I would say that the more options for theming, the better (as long as there is no downside like increased CPU usage). (That said, perhaps the full regex pattern, used to search the code for the syntax elements, could be modified on a theme-by-theme basis, that is, if your chosen theme does not have a color entry for a particular syntax element, then the regex for that element need not be applied.)

I was also considering whether the name of a class in a class definition could also have its own color.
 
> Thonny also gives separate coloring class for built-in names (eg `len`). When they are called, it would get two classes -- "builtin" and "function". Have you considered, which should take precedence, when both are assigned a different foreground color?
> 

I would say that the builtins are a more specific element than function calls, so they should take precedence over them (it seems that this is already the case, although I have not looked into the reason).

> You only considered latin letters for function names. In non-English countries, students are likely to use other letters in names as well. I suggest using `\w` here. I think it's ok to allow the name to start with number here, because number literals can't be followed by an opening paren.
> 
> You named the classes in plural. Other classes are in singular, so I suggest to use singular here as well.

I will fix these.

aivarannamaa:
> I would say that the builtins are a more specific element than function calls, so they should take precedence over them (it seems  that this is already the case, although I have not looked into the reason).

I think it's currently due to ordering of some operations, but it's possible to explicitly order one tag to take precedence over another tag (or all tags). See ` .tag_raise` at https://tkdocs.com/shipman/text-methods.html and https://github.com/thonny/thonny/blob/5fa247a89ac18c3930d5674db9864f1903122310/thonny/plugins/coloring.py#L106

> I was also considering whether the name of a class in a class definition could also have its own color.

Already done: https://github.com/thonny/thonny/blob/5fa247a89ac18c3930d5674db9864f1903122310/thonny/plugins/coloring.py#L185

campbelljc:
> 
> > I was also considering whether the name of a class in a class definition could also have its own color.
> 
> Already done:
> 
> https://github.com/thonny/thonny/blob/5fa247a89ac18c3930d5674db9864f1903122310/thonny/plugins/coloring.py#L185

Right. Sorry, I should have clarified. I was thinking it could be useful to separate the colouring in a header (i.e., after 'def' or 'class') of a classname and function name. For example, in the slides for my class, I use blue for class names, as a reminder of the 'class as blueprint' analogy.

thonny/plugins/coloring.py
```diff
@@ -112,6 +112,7 @@ def _raise_tags(self):
         self.text.tag_raise("open_string3")
         self.text.tag_raise("open_string")
         self.text.tag_raise("sel")
+        self.text.tag_raise("builtin")
```
aivarannamaa, line 115: I think it's better to raise "builtin" over function calls (ie. the two argument variant of `tag_raise`. Tag `sel` (for selected) text should be the topmost, othewise the selection could get illegible foreground.

aivarannamaa:
> I was thinking it could be useful to separate the colouring in a header (i.e., after 'def' or 'class') of a classname and function name.

You are welcome to add this distinction to this PR -- in the location I indicated, leave `"definition"` as it is, but add another tag for the same source range depending on the keyword, eg. `"class_definition"` or `"function_definition"`

thonny/plugins/coloring.py
campbelljc: Ok.

campbelljc:
Ok, I have made the changes.

aivarannamaa:
Thanks! Looking good now.

I will merge this to 3.3 branch myself and release in 3.3.12 (hopefully next week)

PR #1900, opened 2021-07-17
Flatpak

### LINKED ISSUE #968: Flatpak build
Please, consider supporting [Flatpak](http://docs.flatpak.org/en/latest/introduction.html#reasons-to-use-flatpak) and publishing thonny on [Flathub](https://github.com/flathub/flathub/wiki/App-Submission).

jwillikers:
This MR is to add a development Flatpak manifest to to the Thonny repository.
This work will allow developers to build Flatpaks locally with the latest source and makes it easy to integrate into CI workflows if desired.
The work here should be easy to transfer to a repository on Flathub.

At the moment, the Flatpak Pip Generator is having difficulties with a couple of dependencies, so this isn't working yet.
The current issue appears to stem from Pylint 2.8.3 and occurs when flatpak-builder is building the Python dependencies.

Fixes #968

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 *.aps
 *.ipdb
 *.dmg
+.flatpak-builder/
 .venv
 .vs
 .idea
@@ -33,6 +34,7 @@ MANIFEST
 RemoteSystemsTempFiles
 installer_building.log
 xxl_installer_building.log
+packaging/linux/thonny-flatpak-build-dir/
 packaging/windows/freezing.log
 packaging/setuptools/*.whl
 packaging/setuptools/thonny.egg-info
--- a/CREDITS.rst
+++ b/CREDITS.rst
@@ -92,6 +92,7 @@ Source contributors, sponsors, advisors, translators and frequent bug-reporters
 * Jens Diemer
 * Juan Falgueras
 * Jonathan Campbell
+* Jordan Williams
 * jose1711
 * José Carlos García
 * Kauri Raba
--- a/packaging/linux/README.md
+++ b/packaging/linux/README.md
@@ -1,0 +1,57 @@
+# Linux
+
+## Flatpak
+
+### Build
+
+Get the source code.
+
+    git clone https://github.com/thonny/thonny.git
+    cd thonny/packaging/linux
+
+Add the Flathub repository.
+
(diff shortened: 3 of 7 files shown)


jwillikers:
I've figured out a workaround for getting the Python dependencies working.
The development Flatpak appears to be working pretty well.

@aivarannamaa Would you be able to open a PR to the Flathub repository according to the instructions found [here](https://github.com/flathub/flathub/blob/master/CONTRIBUTING.md)?
I've prepared my own fork for you to use [here](https://github.com/jwillikers/flathub/tree/org.thonny.Thonny).

jwillikers:
Plugins appear to work great in the Flatpak.
The main issue I've managed to find is that `tools > Open system shell...` doesn't work correctly since it can't find a terminal command within the Flatpak.
I believe it's possible to accomplish this by checking for available terminals and opening one using the `flatpak-spawn` command from with the Flatpak.
This will require some modification in `terminal.py`
The existence of the `flatpak-spawn` command should reliably indicate that Thonny is being run from inside a Flatpak.

aivarannamaa:
Thank you for this work and especially for the comprehensive readme!

Why did you use `.devel` suffix in the application id? Is it meant to be a test-app or it has another meaning? Would it be OK, if I just remove this suffix?

jwillikers:
> Thank you for this work and especially for the comprehensive readme!
> 
> Why did you use `.devel` suffix in the application id? Is it meant to be a test-app or it has another meaning? Would it be OK, if I just remove this suffix?

@aivarannamaa Great question. I added the `.Devel` suffix to distinguish the development addition of the Flatpak from the release version that will be available on Flathub once you submit there. The suffix isn't necessary by any means, but it does allow installing a development version of the Flatpak alongside a released one. Many GNOME applications follow this convention. [GNOME Podcasts](https://gitlab.gnome.org/World/podcasts/) is a good example. The Flathub Flatpak will track specific releases of Thonny, but the development version built from the Flatpak Manifest here can be used to check the latest changes. You'll see in the Flathub repository I've made for you to copy from, I removed the `.Devel` extension and changed the *source* to come from a specific Thonny release instead of *master*.  I hope that makes sense. Let me know if you have any questions or need help with the submission to Flathub!

PR #1912, opened 2021-07-24
Flatpak fixes

jwillikers:
I found a few things to polish up for the Flatpak.
Namely, I've corrected the AppData file so that it can be accepted on Flathub and hidden the `Open system shell...` menu when Thonny is running from within a Flatpak.
The former required documenting releases of Thonny as part of the AppData file, i.e. the `packaging/linux/org.thonny.Thonny.appdata.xml` file.
This information is conveniently shown to the user in app stores.
It would be good to update this file before cutting a new release with at least the version number and date.

I've also updated my example [Flathub fork for Thonny](https://github.com/jwillikers/flathub/tree/org.thonny.Thonny) which should make it easy to submit the app on Flathub since it contains the necessary files.
Since I updated the AppData file in Thonny's own repository, it will be easiest to use the corrected AppData file when it is available in the next release of Thonny.
When that happens, I'll go ahead and update the version in my fork and then it should be ready for a Thonny maintainer to take the files and submit them to Flathub through their own PR.

### DIFF
--- a/packaging/linux/README.md
+++ b/packaging/linux/README.md
@@ -1,5 +1,12 @@
 # Linux
 
+## Release
+
+On Linux, the AppData file, i.e. `org.thonny.Thonny.appdata.xml` for Thonny, contains information presented users in app stores.
+This includes information about each release.
+Thus, before cutting a release, it's important to add at least the version of the release and the date to the top of the `releases` section of the AppData file.
+It's also helpful to add a description of the changes the new release brings with it.
+
 ## Flatpak
 
 The instructions here describe how to build and update the development version of the Thonny Flatpak.
--- a/packaging/linux/org.thonny.Thonny.Devel.yaml
+++ b/packaging/linux/org.thonny.Thonny.Devel.yaml
@@ -5,6 +5,12 @@ runtime: org.freedesktop.Platform
 runtime-version: "20.08"
 sdk: org.freedesktop.Sdk
 command: thonny
+tags:
+  - devel
+  - development
+  - nightly
+# todo Better distinguish between the release and development installations by using a dedicated icon to represent the development Flatpak instead of prefixing the name.
+desktop-file-name-prefix: "(Development) "
 cleanup:
   - /man
   - /share/man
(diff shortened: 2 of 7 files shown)


aivarannamaa:
Thanks! 

Do I understand correctly, that `flatpak-spawn` command, you mentioned earlier, didn't work out for opening the terminal?

I prefer not to have missing commands in some builds (because it can result in countless bug reports and forum questions), so I restored the terminal-related menu items and made them show an explanation when run inside flatpak.

aivarannamaa:
One more thing. I now noticed, that you pulled packages also from `requirements-xxl-bundle.txt`. Did you do this because it is not possible to install more packages afterwards?

jwillikers:
I got pretty far with getting the `flatpak-spawn` command to work, to the point it could open a terminal outside the Flatpak though it opened two terminal windows instead of one. However, it occurred to me that this completely circumvents the sandboxing Flatpak provides, because the application can do anything it wants on the host system at that point. This is likely to cause some contention when you get around to submitting the Flatpak manifest upstream to Flathub since they don't really like apps having more permissions than necessary in order to protect end users. From what I figure, the best way to open an external terminal session would be to use the [AppChooser portal](https://docs.flatpak.org/en/latest/portal-api-reference.html#gdbus-org.freedesktop.impl.portal.Access), otherwise it might be possible to use some of D-Bus call(s) to the external terminal application(s).

Thanks for making those changes, I'd hate to have the changes cause a bunch of bug / forum traffic.

As for the `requirements-xxl-bundle.txt`, I chose to bundle those with the Flatpak to avoid issues with users installing them inside the Flatpak. When something is installed within the Flatpak, it will be erased when the user's installation is updated to a newer version. To avoid this, I just bundled them all directly.

This isn't a problem with plugins since they are installed outside the Flatpak in `~/.local/lib`, which isn't exactly preferred since both a Flatpak and non-Flatpak version of Thonny might share this directory, and Flatpak's are generally supposed to be self-contained. This last item is something I could look into working around using Flatpak's `--persist` option, but I digress.

PR #1917, opened 2021-07-28
Isolate the user's site packages for the Flatpak

jwillikers:
Currently, the Thonny Flatpak uses the user's site packages directory, `~/.local/lib`, on the host system.
Flatpaks should be isolated from alternative installations of the same application on a system.
If Thonny is installed via the system package manager and as a Flatpak, both versions will share the same plugins.
This PR configures the Flatpak installation to use it's own dedicated, persistent directory for Python site packages.
I also removed a permission which doesn't need to be specified explicitly.

### DIFF
--- a/packaging/linux/org.thonny.Thonny.Devel.yaml
+++ b/packaging/linux/org.thonny.Thonny.Devel.yaml
@@ -25,7 +25,6 @@ finish-args:
   - --share=ipc
   - --share=network
   - --socket=x11
-  - --talk-name=org.freedesktop.FileManager1
 modules:
   - name: tcl
     buildsystem: autotools
--- a/packaging/linux/org.thonny.Thonny.Devel.yaml
+++ b/packaging/linux/org.thonny.Thonny.Devel.yaml
@@ -19,8 +19,8 @@ finish-args:
   # Thonny requires access to serial / USB devices for embedded development.
   - --device=all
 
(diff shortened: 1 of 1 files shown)


aivarannamaa:
Thanks! I'll test and make the PR against flathub repo tonight!

BTW, I see that you have configured Thonny to run with Python 3.8. Now that Debian Bullseye is about to be released with Python 3.9, Raspberry Pi will also move to Python 3.9. It would be nice if flatpacked Thonny would also contain Python 3.9. Would it be easy to change this?

jwillikers:
I tinkered with using 3.9 a little bit when I was first creating the Flatpak, and it might be possible if the same version of Python is used to generate the pip dependencies via the `flatpak-pip-generator` as is used in the Flatpak. Using different versions seemed to lead to some issues with that tool. I decided to play it safe and match the Python version in the Freedesktop Platform /Sdk version 20.08, which is used as the base system. The [Freedesktop Sdk](https://gitlab.com/freedesktop-sdk/freedesktop-sdk) is updated annually in August, so I think we'll get Python 3.9 when the Freedesktop Platform 21.08 becomes available. At that point, I think it will be easy so far as the `flatpak-pip-generator` doesn't give me any trouble.

Sounds good. Let me know if you have any trouble testing it. I've updated [my Flathub fork](https://github.com/jwillikers/flathub/tree/org.thonny.Thonny) with the files to be submitted to Flathub.

I'll also submit a quick PR here to add the latest release to the AppData file.

PR #1922, opened 2021-07-31
Mirror what will be on Flathub

jwillikers:
This just includes the minimal number of dependencies at the moment.
It's possible to include more, but becomes a larger maintenance burden.
This commit also forgoes reinstalling all of Python for tkinter.
It now just installs the tkinter module itself.

See PR flathub/flathub#2442.


jwillikers:
I've been able to re-package everything so far in the Flatpak except for Pygame in my fork. I don't see any Flatpaks on Flathub shipping anything newer than 1.9.4, which is the latest version packaged in the flathub/shared-modules repository, but it looks like we need `2.0.*`.

aivarannamaa:
TBH, for me it feels weird to include so many extra packages in the flatpak, considering how easy it is to add packages in Thonny (at least on the main platforms -- I don't know the situation with binary wheels for Raspberry).

(Yes, Thonny's XXL bundles also contain many packages, but these bundles exist, because Raspberry Pi really wanted to have this selection of packages to be readily available for other platforms besides Raspberry and I gave in.)

Can you think of anything related to package management, which could be changed in Thonny so that you wouldn't want to include many packages in Thonny's flatpak anymore?

jwillikers:
Awww, I see.
I thought there might be a large number of users expecting to have the XXL packages installed, which would make having those installed lead to more reproducible builds when doing bug fixes and such.
I'm always for anything that reduces the maintenance overhead, and it's easy to cut out anything we don't want to bundle from the Flatpak manifest.

I don't think packages are installed under `~/.local/lib` like plugins are from a quick test. When I attempt to install Black as a *system package*, I get an error that Pip is not installed. It's possible to workaround this using a virtual environment. Python packages which **do not** require any specific system packages, i.e. LAPACK for SciPy and SDL for PyGame, should be pretty safe and easy to install this way. I think I saw you have an updater that would take care of updates between minor Python versions, which should help too.

As for the the packages relying on software on the base system, this might have some strange behavior when the base system updates but the installed Python package remains. If a user installs something like PyGame, it has to install SDL. SDL should be installed just fine using the Python wheel, right? If SDL is installed outside of `~/.local/lib` and any other persistent user directories such as `~/.config`, `~/.cache`, and `~/.local/share`, then there might be problems when the Flatpak updates to a newer version. SDL might disappear while Pygame remains installed, which could require completely reinstalling the package. As I'm a C++ developer by trade, my knowledge of Python packaging formats is a bit lacking, so I'm not sure where the Python wheel would install something like SDL.

I'm happy to go with bundling less packages, and we could always for some subset of packages. For now, I'll go ahead and assume we at least want to bundle the *regular* packages and submit a PR accordingly. I'll make sure that my work at installing the remaining packages is available in the commit history for easy access.

jwillikers:
Perhaps all that's necessary is a way to inform the user to install packages in a virtual environment when installing *system packages* in the Flatpak, or maybe we just install to the user's site-packages by default in the Flatpak as is done for plugins? If having to reinstall packages becomes an issue, we can always notify users / reinstall packages for them, perhaps? Those are my ideas at the moment.

jwillikers:
Oh wait, I just realized that switching to the proper way of installing Python packages happens to leave out Pip, so plugins can't be installed at the moment unless a virtual environment is used.

jwillikers:
Switching the runtime to the Sdk should allow installation like normal.

PR #1924, opened 2021-08-01
Bundle the standard set of packages in the Flatpak

jwillikers:
See the discussion started in #1922 about which packages to bundle in Thonny.

This PR adds the standard set of packages bundled with the Thonny distribution to the Flatpak.
The Git history includes logic for bundling all of the XXL packages, so that these can be included in the future if desired.


jwillikers:
Okay, so obviously I've gotten a case of *dependency* brain and have already bundled the standard set of packages. This just adds the XXL packages in the commit history and does a bit of cleanup. We can further discuss / change what packages to pull in here, if needed.

jwillikers:
I've updated my Flathub fork to use the standard set of packages to reflect our discussion and what's in this repository's development Flatpak manifest.

jwillikers:
One second too, slow :)

Issue #66, opened 2015-04-08
Name refactor - renaming to existing filename


aivarannamaa:
**[Original report](https://bitbucket.org/bitbucket-issue-migration\thonny-issues.zip/issue/66) by Anonymous.**

----------------------------------------

When a file A is renamed via 'name refactoring' to B but a file C with the same name already exists, then this is currently allowed and the existing file is just overwritten. Nothing that can or should be done about that currently. 

But if the user already had file C open, then what happens now is there are two tabs open, both named B, and having the same content. This creates all sorts of issues. 

Best fix is to probably implement a Thonny-wide mechanism for not allowing two tabs referring to the same file to be open, if this situation occurs then just close one of the tabs automatically.

Issue #1903, opened 2021-07-19
Go to line number.


jrs69:
I'm new to Thonny, all good!, except I would love a shortcut to go to a specific line, e.g. ctrl+g or ctrl-l etc.

I'm a retired programmer just doing hobby stuff, but were I still working it would be a problem.

aivarannamaa:
Thank you for the suggestion!

This would be a nice and non-obtrusive addition. I'll add it into my TODO list.

Related to #1665

neil9:
Hear hear. It would be VERY nice to be able to go directly to the line number rather than having to scroll through the whole code.

bigrom26:
I try go to selected line but not work Ctrl-g and not in menu. 
How correct this problem ? 
I work on Win10.

aivarannamaa:
@bigrom26, this feature isn't implemented yet.

aivarannamaa:
Implemented in #2001

PR #2001, opened 2021-10-16
usability features

kr-g:
please review.
 
this is also related to https://github.com/thonny/thonny/issues/66


thonny/base_file_browser.py
```diff
@@ -863,12 +880,132 @@ def should_open_name_in_thonny(self, name):
...
+        return len(self.paths) > 0
+
+    def dirname(self, path):
+        if os.path.isfile(path):
+            path = os.path.dirname(path)
+        return path
+
+    def conflicts(self, target):
+        target = self.dirname(target)
+        folder = set(map(lambda x: self.dirname(x), self.paths))
```
aivarannamaa, line 961: Do I understand properly, that you want to avoid the situation where any of the source files gets pasted to where it already is? 

I think this is valid, but you should also forbid cut-pasting into a directory which is _below_ any of the selected items.

thonny/base_file_browser.py
```diff
@@ -703,6 +706,20 @@ def add_middle_menu_items(self, context):
...
+
+        if self.supports_rename():
+            self.menu.add_command(label=tr("Rename"), command=self.rename_file)
+
+        if self.supports_copypaste():
+            self.menu.add_command(label=tr("Copy"), command=self.copy_files)
+            self.menu.add_command(label=tr("Cut"), command=self.cut_files)
+            if self.copypaste.has_selection():
+                target = self.get_selected_file()
+                if target and not self.copypaste.conflicts(target):
```
aivarannamaa, line 720: What do you think of omitting `conflicts`-check here and relying on the check below? I'm not sure the user will always understand what is the problem when the "Paste" command is simply missing.

thonny/editors.py
```diff
@@ -905,6 +917,13 @@ def close_single_untitled_unmodified_editor(self):
         if len(editors) == 1 and not editors[0].is_modified() and not editors[0].get_filename():
             self._cmd_close_file()
 
+    def _cmd_goto_source_line(self):
+        editor = self.get_current_editor()
+        if editor:
+            line_no = simpledialog.askinteger(tr("Goto"), tr("Line number"))
+            if line_no:
+                editor.select_line(line_no)
```
aivarannamaa, line 925: Have you tested that too big and negative integers don't crash?

aivarannamaa (changes requested):
Thank you for this work! I'm really sorry for so late feedback!

I like what you did and I'm ready to include this in Thonny 4.0, but please consider my suggestions first.

thonny/editors.py
kr-g: thonny's implementation already works with negative and too big numbers.
so i have not added an extra check.

aivarannamaa:
Are you going to add the extra conflict check for cut-paste? Or do you maybe disagree with my analysis?

kr-g:
for me it is not fully clear what "below" means.
assuming following folder structure where file1 ist the source, and future_place_of_file1(opt-a/b) is the destination.

someplace-+
                   +-projectbase-+
                   |                  +-src-+
                   |                           +-file1
                   |                   +-tests-+
                   |                               +-future_place_of_file1(opt-a)
                  +-setup.py
                  +-future_place_of_file1(opt-b)

where in this picture is your "below" ?
or is "below" a piece of code fragment what i missed to see here in this conversation so far???
(i mean .... could be my mistake...)

kr-g:
have you seen this ?
https://github.com/thonny/thonny/pull/2001/commits/2110564e37fdee4cbc3721bb2444c30c38d813d9

aivarannamaa:
What I meant is that the user shouldn't be able to cut `projectbase` and paste it under `tests`.

kr-g:
true, this would be best - inspecting all paths of all selected first.
thats also what i meant in an earlier post (wrt pros and cons of shutil)

btw. in linux file browser (thunar) if copies the following

Bildschirmfoto_2021-11-08_12-28-45

shutil never stops (i was surprised to see...)

Bildschirmfoto_2021-11-08_12-31-18

kr-g:
ok, i will have a proper look at it (probably) tomorrow

kr-g:
commit 
https://github.com/thonny/thonny/pull/2001/commits/0341939f50dddf0dbcf4490d2cc0e1b6b7b5a894
is related to 
https://github.com/thonny/thonny/issues/1903

kr-g:
enable and disable of "paste" menu seams to work 
dont know what shutils does in the several involved use-cases

aivarannamaa:
Are you done with the cut/copy-paste stuff?

Regarding remembering line -- it seems to be unrelated to copy-paste, so please submit it as a separate PR. Also, I'd prefer if this specific case of remembering the line goes without a configuration option.

kr-g:
i m done for now.
i also dont have much time in the next time contributing to thonny.
so it would be great if you could form the code to your needs whilst merging.
sorry for that.
next time i will create an own branch rather than working on master.

aivarannamaa:
OK. Thank you for your work!

PR #2026, opened 2021-10-30
Update German translation

iblech:
Looked through lines 1 to 1000. Source: am native German speaker.

Thank you for your work on Thonny, I'll use Thonny this Monday for a Python tutorial for beginning university students. :-)


aivarannamaa:
Thanks! I'll merge this for now, but so far we have managed translations via POEditor (https://github.com/thonny/thonny/blob/master/docs/translate.md#short-story-for-translators). This policy may change in the future, though.

I'll import it in POEditor myself.

I'm glad you've found Thonny useful!

iblech:
Oh, I didn't notice that. Thank you for letting me know and for importing it in POEeditor! For the next batch of translation updates, I'll use POEeditor.

Issue #1644, opened 2021-02-22
Support loading environment variables from file


aivarannamaa:
see https://code.visualstudio.com/docs/python/environments#_environment-variable-definitions-file

PR #2049, opened 2021-11-16
loading environment variables from file

kr-g:
this belongs to 
https://github.com/thonny/thonny/issues/1644


kr-g:
gosh!
seams that all of them are again in the pull request.
not only from my branch :-(

kr-g:
ok i will wait for your reply if this causes too much trouble by merging.
otherwise i will kick the PR and clone thonny freshly and redone the few lines...

btw.
the python-dotenv module introduced in requirements.txt can be found here
https://github.com/theskumar/python-dotenv/tree/master/src/dotenv

kr-g:
another thing.
the Variable substitution mentioned in 
https://code.visualstudio.com/docs/python/environments#_environment-variable-definitions-file

is not yet supported by python-dotenv.
but i guess that's fine at the moment.
one of us could create an enhancement request there if it's of a need...

kr-g:
add.remarks.

in get_environment_with_overrides() PATH is updated when given with update_system_path(),
and variables are droped when None.

should this also be supported?
i think removing first (existing thonny implementation), and then just adding (with .env file) is more intuitiv 

RFC

kr-g:
just created 
https://github.com/theskumar/python-dotenv/issues/365
for variable substitution

aivarannamaa:
Sorry for the late reply!

I'm hesitant to introduce another dependency, especially if it's not very useful for Thonny's target group (the beginners).

What do you think of avoiding the library and implementing the required part yourself? As I see it, without variable substitution it's simply a for loop over a file, and splitting each line on the first `=`.

kr-g:
the mentioned module offers some neat features - but ... i can do. why not.
in principle it is like you say: just spliting at the first "="
and i think comments introduced with "#" at the start, or end of line should be supported too.
ok i will rollback the PR and continue with own coding...

Issue #1842, opened 2021-05-26
Create the dialog for specifying launch configuration


aivarannamaa:
Related to #1087 and #1121

The plan is to maintain a single configuration per interpreter kind (CPython or Unix MicroPython vs bare metal MicroPython/CircuitPython). Named configurations (a la PyCharm) could be useful for advanced users, but for beginners they would simply add another obstacle (creating a configuration) for getting things done.

The configuration will be managed via a dialog with initial content something like this (for CPython):

```
Main file or module: (*) current file   ( ) specific file   ( ) specific module
Working directory:   ( ) current directory   (*) directory of the main file   ( ) specific directory
Program arguments:   (*) no arguments  ( ) one or more arguments separated by spaces
Program inputs:      (*) provided via shell  ( ) provided here, one input per line
Program output:      (*) write to Shell  ( ) write to file
Environment values:  (*) default  ( ) add or override here, one NAME=VALUE per line
```

In each radio-button group there is one or more options (eg "specific file"), which cause a text box or editable combobox (possibly with file or directory selection button) to be folded out under corresponding setting:

```
Main file or module: ( ) current file   (*) specific file   ( ) specific module
                     [ C:\Users\Aivar\School\programming\blah-blah.py                                v] [...]
Working directory:   ( ) current directory   (*) directory of the main file   ( ) specific directory
Program arguments:   (*) no arguments  ( ) one or more arguments separated by spaces
Program inputs:      ( ) provided via shell  (*) provided here, one input per line
                     [ 345                                                                            ] 
                     [ yes                                                                            ] 
Program output:      (*) write to Shell  ( ) write to file
Environment values:  (*) default  ( ) add or override here, one NAME=VALUE per line
```

The dialog will be non-modal, ie. it can be left open if you want to quickly test your program with several sets of arguments or inputs. It can be opened with a new button in the right end of the toolbar (the button text will contain a summary of the current configuration). Also, to maintain backward compatibility, current "View => Program arguments" will also open this.

aivarannamaa:
New plan

### Local

```
Main file:           (*) current file   ( ) specific file 
Run as:              (*) file   ( ) module (-m)   ( ) unnamed code (-c)
Working directory:   ( ) current directory   (*) directory of the main file   ( ) specific directory
Program arguments:   (*) no arguments  ( ) one or more arguments separated by spaces
Program inputs:      (*) provided via shell  ( ) provided here, one input per line
```

[...]

Issue #1755, opened 2021-04-07
Make MP back-end advise Ctrl+C even if only whitespace characters get printed


bitboy85:
If the device is already executing code thonny does not show its files or some output in the commandline.
The os filemanager still allows editing the files on the device.

I think it would be helpful to print a line like "The device is currently running. Press CTRL + C to abort" or something like this in the comandline line.

Firmware: Adafruit CircuitPython 6.2.0-beta.3-178-gd2563c57d on 2021-03-13; Raspberry Pi Pico with rp2040

aivarannamaa:
Thank you for the suggestion!

Did you get anything in the output when your device was running and you connected Thonny to it? Actually Thonny is supposed to print this information, but only if the device doesn't print anything in 2.5 seconds: https://github.com/thonny/thonny/blob/707e69ec3a567df5f82205c5a2ae0d79f186ed25/thonny/plugins/micropython/bare_metal_backend.py#L820

bitboy85:
The Output is just an empty line.
Hitting Enter will give you an error message "WARNING:root:Unexpected echo. Expected b'\r\n', got b''"
![grafik](https://user-images.githubusercontent.com/6197580/113988698-03978780-9850-11eb-8ce6-5d7a89d9d969.png)

aivarannamaa:
Ok, I should then consider only visible characters when deciding whether to print the advice or not.

aivarannamaa:
Thonny 4.0 will print the "device is busy or doesn't respond" message when no visible chars appear after certain time after connection or when interrupt doesn't have an effect.

PR #2060, opened 2021-11-25
loading environment variables from file

kr-g:
as proposed in https://github.com/thonny/thonny/pull/2049
except replace variables, or setting path


kr-g:
remark:
build failed because former build also failed

cause:
https://github.com/thonny/thonny/issues/1755

kr-g:
i did a first draft for expanding vars but i lack of time to test properly and integrate it

https://gist.github.com/kr-g/c5971bffa2ae18b2de1b2aa6e69c364e

thonny/running.py
```diff
@@ -865,6 +865,64 @@ def open_custom_system_shell(self):
         raise NotImplementedError()
 
 
+class EnvFile(object):
+    def __init__(self, fnam=None):
+        self.fnam = fnam if fnam else ".env"
```
aivarannamaa, line 870: Why not simply use `".env"` as the default value?

thonny/running.py
```diff
@@ -916,8 +974,42 @@ def _get_environment(self):
...
+        # un-set keys
+        env = EnvFile()
+        env.set_env(chained_env)
+        env.remove_unused()
+
+        return env.get_env()
+
+    def _load_env_file(self):
+        # todo: configuration menu -> Tools > Options > Run & Debug
+        env_file = os.environ.get("THONNY_ENV_FILE", ".env")
```
aivarannamaa, line 992: THONNY_ENV_FILE must be simply filename without path, right?

thonny/running.py
```diff
@@ -916,8 +974,42 @@ def _get_environment(self):
...
+                if env.excep:
+                    logger.exception("env-file" + str(env.excep))
+            env = env.get_env()
+        else:
+            env = {}
+
         return env
 
+    def _get_launcher_python_script_dir(self):
+        return self._get_launcher_with_args()[-1]
```
aivarannamaa, line 1011: Here you are simply returning last argument -- this doesn't look right.

I would look for the `.env` from the working directory of the backend process. See the first option in "Tools => Options => Run & Debug"

thonny/running.py
```diff
@@ -916,8 +974,42 @@ def _get_environment(self):
...
+
+        add_env_file = self._load_env_file()
+
+        chained_env = collections.ChainMap(add_env_file, env)
+        chained_env = dict(chained_env)
+
+        # un-set keys
+        env = EnvFile()
+        env.set_env(chained_env)
+        env.remove_unused()
```
aivarannamaa, line 986: I see you are treating the empty value as signal to remove the environment variable. Is this standard behavior? Why not simply keep the variable with empty value?

aivarannamaa (changes requested):

thonny/running.py
kr-g: you mean like ?
def __init__(self, fnam=".env"):

pep 671 status is "draft" 
https://www.python.org/dev/peps/pep-0671/

so i would propose to add an comment
# todo pep 671

thonny/running.py
kr-g: no. 
it also can be a absolute, or relative path.

refer to [this](https://github.com/thonny/thonny/blob/6e7b9cf572a46cd95bdd38a5411064b9649d8b7f/thonny/running.py#L881) where it is just used with open.

thonny/running.py
kr-g: i got this by putting a print statement into [running.py -> _start_background_process](https://github.com/thonny/thonny/blob/6e7b9cf572a46cd95bdd38a5411064b9649d8b7f/thonny/running.py#L1013)
and checked what happens when i press "run" button in thonny.

but since i was also quite happy with my findings i put it into an own function because i think that sort of function should be available e.g. `get_workbench().get_launcher_python_script_dir()`

thonny/running.py
kr-g: a good point!

but how then to support not existence when falling back to the programmers default with `os.environ.get("somekey", defaultval)` is the wanted behaviour ?

how about adding an additional hard test against double and single quoted strings to get this feature for empty strings.

or how about to stay with that implementation and wait if issues are created asking for such a feature, where someone has to point out why it is required. thinking here of a step-by-step process to add more features. also because ".env" file settings and pythons configparser functionallity might overlap here.

thonny/running.py
kr-g: oops. small correction. i just went too quickly thru my own coding...
its used with path join - so it is just the filename without path.

thonny/running.py
kr-g: with a function encapsulating `env_file = os.environ.get("THONNY_ENV_FILE", ".env")` so that the strig ".env" does not appear at several places

aivarannamaa: As far as I can see, on this line pep 671 is not relevant, as the default value is fixed and does not need to be late bound. Anyway, this is a minor issue, so feel free to keep it.

thonny/running.py
aivarannamaa: I see, I had forgotten, that _get_launcher_with_args puts working directory as the last argument.

IMO it would be more direct to use `self.get_cwd()`.

Also, I now noticed that you added stuff to the SubprocessProxy, but I think it should be in a more specific class -- CPythonProxy. SubprocessProxy is used also with MicroPython and Ssh backends, and the .env file is not relevant there (at least not directly). For example, if you select "Remote Python 3" backend, then you probably want the .env file to be applied only on the remote host.

thonny/running.py
aivarannamaa: I think this should work just like in other dotenv implementations (in case there is no specification which clears this issue).

kr-g:
Ok. Thanks for the feedback so far.
I will rework within the next week.

thonny/running.py
kr-g: probably ".env" should be named as "thonny.env" instead;
since a dot-file is a hidden file under some systems e.g. raspberry pi.
just to apply to thonny's idea that it should be easy to understandable by beginners.
and also because hidden files are not shown by default.
but you decide.

thonny/running.py
kr-g: there might be a specification as implementation (aka. living standard by dotenv module itself)
i keep up with having bash var behaviour in mind. 
in this case treated as empty string.

kr-g:
added feature for expanding vars 

also supported overlaying e.g.
   PATH = ${PATH}:some_other_path
works...

thonny/running.py
kr-g: i rebased my branch to master.
regarding SubprocessProxy i will have to do more investigations... 

one maybe stupid question here:
why not having sort of propagation of env settings here to MicroPython and Ssh backends?
so that it is also available there like you mentioned already... 

> For example, if you select "Remote Python 3" backend, then you probably want the .env file to be applied only on the remote host.

thonny/running.py
kr-g: changed the implemenation so that no default is used anymore here

thonny/running.py
aivarannamaa: I'd go for `.env` as I don't think dotenv is begginner's topic.

thonny/running.py
aivarannamaa: > why not having sort of propagation of env settings here to MicroPython and Ssh backends?
> so that it is also available there like you mentioned already

With SSH back-end it wouldn't propagate anyway over SSH connection -- the variables need to be set explicitly on remote host (https://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.exec_command). And it wouldn't be harmless either -- in principle a variable that is fit for the remote process may mess up the local process.

thonny/running.py
kr-g: Ok

thonny/running.py
kr-g: The local env is not touched by this implementation.
Or? 

So with the given link to exec_command documentation the propagation should be easy to implement.

How do we deal with this now?
I m asking because I can not do much more this year, and I m somehow unaware of thonny 4.0 release timeline...

So one possibility could be merge current code and open a new issue regarding the SSH limitation and propagation Todo.

So... ???

aivarannamaa: > The local env is not touched by this implementation.

Currently theres two local processes when you're using ssh backend -- the UI process and another local process, which communicates with the remote host. I'm may merge the two for 4.0, though.

I hope to get the first beta of 4.0 out before the end of December. If you can't proceed by then, then I guess I'll merge the current state and fix it as I see fit.

kr-g: Probably I have some time between Xmas and new year. 
Otherwise just merge and create an issue, or add the missing SSH integration code

kr-g:
merged latest thonny-master and resolved conflicts

kr-g:
i had no look into this since a while now.
please do let me know regarding:

> Currently theres two local processes when you're using ssh backend -- the UI process and another local process, which communicates with the remote host. I may merge the two for 4.0, though.

aivarannamaa:
> Currently theres two local processes when you're using ssh backend -- the UI process and another local process, which communicates with the remote host. I may merge the two for 4.0, though.

I've decided to postpone this merge at least until 4.1

kr-g:
https://github.com/thonny/thonny/issues/1842

kr-g:
one note on integration of remote / shh.
since the remote `os.environ` is unknown in thonny when running a single file via ssh there should be a process to get the `os.environ`, and merge this with the thonny local `.env` file before running the remote script
this could be achieved for instance by running `set` in bash / cmd.exe what will return the remote `os.environ`. for performance reasons this could be cached after the initial ssh connection was establish / defined (assuming that the remote `os.environ` will not change frequently... having in mind to reset the cache on backend restart)
remember to start bash in `--posix` mode to prevent line breaks in `set` output
RFC

kr-g:
one more...
`/bin/sh` would be the better shell to use - since it's always available

aivarannamaa:
Thanks for the updates! I'll deal with this topic after I've had time to fix more bugs in Thonny 4.0.0.

kr-g:
some notes on paramiko and handling env variables

hints in official doc regarding paramiko ignoring setting env variables:
https://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.exec_command
https://docs.paramiko.org/en/stable/api/channel.html#paramiko.channel.Channel.set_environment_variable

here a issue regarding this
https://github.com/paramiko/paramiko/issues/936

and in issue#936 in the comments a solution:
`-> in target server, add new line "AcceptEnv MYENV" to /etc/ssh/sshd_config, restart sshd service`

but this is then limited to the listed variables in the config file then (in this case MYENV)
each env var needs to be enabled separately in the config file.

following up here 
https://www.man7.org/linux/man-pages/man5/sshd_config.5.html

-> SetEnv 
--> AcceptEnv
--> PermitUserEnvironment

looks now more complicated than expected before (i have to admit that...)

kr-g:
your wrote:
`And it wouldn't be harmless either -- in principle a variable that is fit for the remote process may mess up the local process.`

i agree.

Issue #2379, opened 2022-08-07
key shortcut for menu "tools -> open system shell..."


kr-g:
can we have a key shortcut for menu "tools -> open system shell..." ???
e.g. Shift+Control+T

aivarannamaa:
Thank you for the suggestion, but I'd rather not fix the use of this shortcut yet -- I may need it for another feature in the future.

But you can add it for yourself by editing Thonny's configuration file: https://github.com/thonny/thonny/wiki/Custom-shortcuts

kr-g:
i think opening a shell is more or less on the same "level" as "run current script in terminal".
so i think it should come also directly out of the box.
but ... yes... if there is another option ...

kr-g:
https://github.com/thonny/thonny/pull/2380#issuecomment-1209225221

i cleaned it from env related code.
sorry again for that.
i dont wanted to waste your time.

kr-g:
https://github.com/thonny/thonny/compare/master...kr-g:thonny:kr-g-terminal-short-key?expand=1

aivarannamaa:
In Thonny 4.1 I want to make it possible to add or change shortcuts in the options dialog (#306)

PR #2380, opened 2022-08-07
open system shell  short key

kr-g:
this is regarding

https://github.com/thonny/thonny/issues/2379


aivarannamaa:
See my reply under https://github.com/thonny/thonny/issues/2379

Also, I wouldn't accept this PR even if I wanted to include this change, because it has unrelated changes next to it. Each feature should be submitted from a separate branch.

kr-g:
yes true, i already cleaned up my repo yesterday evening because somehow i updated the branch with github some long time ago from the master (what was wrong, since that makes of course no sense on a branch, but github offered that button.. and ... i always press buttons...) ... or i dont know why this happend. mostlikely its also already since the begining where i worked directly on the master of my repo...

Issue #2377, opened 2022-08-05
venv backend change in file browser context menu


kr-g:
this is not a bug.

at the present time switching to an alternative backend is done via menu "Run -> Select Interpreter", then choosing a python executable-

background:
when creating a venv a file [`pyvenv.cfg`](https://docs.python.org/3/library/venv.html) is placed in the corresponding folder. this file is a defacto standard of identifying a venv. 
here the default python executable is available via path `./bin/python` 

proposal:
1. when right-clicking on a file named `pyvenv.cfg` the context menu in the file browser contains an entry (e.g. "switch to venv") for switching the option "run.backend_name" for the workbench. like in class [`BackendConfigurationPage.apply()`](https://github.com/kr-g/thonny/blob/41dad8afcbbf831a8739ed6f4706ae9f0c344170/thonny/plugins/backend_config_page.py#L152)
2. when right clicking a tree node in file browser with type "folder" plus having a file `pyvenv.cfg` inside the folder the context menu entry should also appear.

RFC

kr-g:
please review before i create a pull request

https://github.com/kr-g/thonny/tree/kr-g-switch-venv

kr-g:
one file changed

https://github.com/kr-g/thonny/commit/111a2ea99238c1f573998cda86d93979b00fca4f

kr-g:
how about an info text on the tab "Interpreter options" describing this ?

kr-g:
i did also the info text now...

RFC

aivarannamaa:
Thank you for the suggestion!

I like this idea! Added a comment under https://github.com/kr-g/thonny/commit/111a2ea99238c1f573998cda86d93979b00fca4f

aivarannamaa:
Thank you!

I made some changes and the feature will be in the next beta.

PR #2381, opened 2022-08-09
switch venv via file browser context menu

kr-g:
this belongs to 

https://github.com/thonny/thonny/issues/2377


kr-g:
cleaned PR from unrelated source

RFC

PR #2389, opened 2022-08-16
fix info text

kr-g:
caused an exception
since checking for wrong content at this point


andypiper:
There's a lot of commits in this PR for a very small change to a single file, maybe consider squashing?

kr-g:
thanks for the feedback.

i will consider to do so next time... 
or probably setup up a new clone - because longer time ago i worked directly on the master (instead on a branch) what introduced some of the entries.

i found this hint (-> since squash is an interactive rebase, https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)

`Don’t include any commit you’ve already pushed to a central server — doing so will confuse other developers by providing an alternate version of the same change.
`

i would leave it therefore as it is (for now, this time...) 

RFC

aivarannamaa:
Thanks!

kr-g:
oh wow, i deleted the branch - but merging was still possible.
good to know.

PR #2490, opened 2022-10-05
Added extra import from logging for logging.error which was used on line 110

solsword:
A student of mine ran into an internal thonny error where every time she tried to run her code, a NameError would come up because logging wasn't defined on line 110 of the mypy plugin's __init__.py. This change fixes that error by adding an import.

The else branch where the error happens must be hard to reach, because only one student out of a class of ~80 had this bug; I was not able to identify what about her code actually triggered that else, although the one thing that was a bit off was that she had imported the current file within itself.

I ran black on this as well as pytest thonny/tests; I hope it's a small and simple enough change that it doesn't break anything :)

### DIFF
--- a/thonny/plugins/mypy/__init__.py
+++ b/thonny/plugins/mypy/__init__.py
@@ -1,7 +1,7 @@
 import os.path
 import re
 import subprocess
-from logging import getLogger, error as log_error
+from logging import getLogger
 from typing import Iterable
 
 from thonny import get_runner, get_workbench, ui_utils
@@ -106,7 +106,7 @@ class MyPyAnalyzer(SubprocessProgramAnalyzer):
                 atts["symbol"] = "mypy-" + atts["kind"]
                 warnings.append(atts)
             else:
-                log_error("Can't parse MyPy line: " + line.strip())
+                logger.error("Can't parse MyPy line: " + line.strip())
 
         self.completion_handler(self, warnings)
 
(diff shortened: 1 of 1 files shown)


kr-g:
refer also to comments in https://github.com/thonny/thonny/issues/2491

kr-g:
@aivarannamaa 
please review
thanks

solsword:
Thanks for spotting that! I didn't see the variable logger although if I'd had more time to look at the code I would have. I've updated the branch to just use `logger.error` instead of having the extra import.

I can confirm that I don't get the internal error popup with this fix, and that it does log an error to the console about the inability to parse the MyPY output line. I gues the MyPY message warning about a duplicate module looks a bit different than the standard MyPY messages it's expecting to see :)

In case it's useful in the future, here are the relevant lines that get logged now that the logger.error call is fixed:

```txt
13:39:02.199 ERROR   thonny.plugins.mypy: Can't parse MyPy line: /Users/pmawhort/tmp/mypy_plugin_test.py: error: Duplicate module named "mypy_plugin_test" (also at "/Users/pmawhort/tmp/mypy_plugin_test.py")
13:39:02.199 ERROR   thonny.plugins.mypy: Can't parse MyPy line: /Users/pmawhort/tmp/mypy_plugin_test.py: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
13:39:02.199 ERROR   thonny.plugins.mypy: Can't parse MyPy line: /Users/pmawhort/tmp/mypy_plugin_test.py: note: Common resolutions include: a) using `--exclude` to avoid checking one of them, b) adding `__init__.py` somewhere, c) using `--explicit-package-bases` or adjusting MYPYPATH
```

Issue #2492, opened 2022-10-06
Add a signature when running applications in Thonny IDE


jmparatte:
Hi,
I'm concerned with identify when running applications in Thonny IDE.
For example when running applications interacting with console input(), to switch input() to something like reading a file.

I propose inserting a line in file **C:\Thonny\Lib\site-packages\thonny\plugins\micropython\mp_back.py** like that:
`         self._prepare_rtc()` **line 157**
`         self._execute("THONNY_IDE={'name': 'Thonny', 'version': '4.0.1'}") # Thonny IDE signature` **inserted line**
`         self._send_ready_message()` **line 158**
So, I can check if running with Thonny IDE like that:
`    try:`
`        if THONNY_IDE: pass`
`        # Thonny IDE detected`
`        ...` 
`    except:`
`        # Thonny IDE not detected`
`        ...`

jmP

kr-g:
i m not sure if i understood your requirement, or use-case, properly.

but probably checking in your code against the platform, or implementation, solves this better together with that `try` block.

with micropython
```
>>> import sys
>>> sys.platform
'esp8266'
```

with linux
```
>>> import sys
>>> sys.platform
'linux'

```

another option cloud be checking against this

```
import sys
sys.implementation

```
or

```
>>> import os
>>> os.uname()

```

please note that not all of the above is available on micropython.
you need to try what is there and fits best.

jmparatte:
Hi, thanks for your reply, but my request is about running in or out of Thonny, to take some special actions if Thonny detected, not to check the implementation or platform, not to check Python or Micropython.

kr-g:
run this
```
import os
for k, v in os.environ.items():
    print(k, v)

```

at the end you will find

```
THONNY_USER_DIR /home/benutzer/repo/thonny/.venv/.thonny
THONNY_FRONTEND_SYS_PATH ['/home/benutzer/repo/thonny', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/home/benutzer/repo/thonny/.venv/lib/python3.8/site-packages', '/home/benutzer/repo/thonny/.venv/.thonny/plugins/lib/python3.8/site-packages']
THONNY_LANGUAGE en_US

```

thonny sets those variables, otherwise they will be not there.
works? fine?

kr-g:
@aivarannamaa 

how about having an addtional env var `THONNY_VERSION` ?

[...]

PR #2494, opened 2022-10-07
env var thonny version

kr-g:
this belongs to 
https://github.com/thonny/thonny/issues/2492


kr-g:
@aivarannamaa 
please review, and pull if requirement is fine with you

kr-g:
thonny `get_version()` method is used at several places, so i changed the handling to return a cached value rather then using a global variable (as proposed with with my initial change)

references 

```
./thonny/__init__.py:def get_version():
./thonny/__init__.py:    main_logger.info("Thonny version: %s", get_version())
./thonny/assistance.py:            "thonny_version": thonny.get_version(),
./thonny/plugins/pip_gui.py:                if req.project_name == "thonny" and thonny.get_version() not in req:
./thonny/plugins/about.py:            self.main_frame, text="Thonny " + thonny.get_version(), font=heading_font
./thonny/plugins/cpython_ssh/cps_back.py:        return f"/tmp/thonny-backend-{thonny.get_version()}-{self._user}"
./thonny/plugins/cpython_ssh/cps_back.py:        if self._get_stat_mode_for_upload(launch_dir) and not thonny.get_version().endswith("-dev"):
./thonny/running.py:    get_version,
./thonny/running.py:        env["THONNY_VERSION"] = get_version()

```

thonny/__init__.py
```diff
@@ -143,11 +143,19 @@ def is_portable():
         return abs_location.startswith("/media/") or abs_location.startswith("/mnt/")
 
 
+_THONNY_VERSION = None
+
+
 def get_version():
+    global _THONNY_VERSION
+    if _THONNY_VERSION:
+        # return a copy
```
aivarannamaa, line 152: Why copy?

aivarannamaa:
Thanks! Looks good now.

PR #2563, opened 2022-11-30
Start command with space to avoid history.

isaacl:
Otherwise, this long string shows up in bash/zsh history!


aivarannamaa:
Thank you and sorry for the delayed response!

PR #2595, opened 2023-01-01
Better i18n support in uf2dialog and cp_front

bletvaska:
* more strings are now available for translation in `uf2dialog` and `cp_front`


thonny/codeview.py
```diff
@@ -246,8 +247,8 @@ def set_content_as_bytes(self, data, keep_undo=False):
                 pass
 
             encoding = ask_string(
-                "Bad encoding",
-                "Could not read as %s text.\nYou could try another encoding" % encoding,
+                tr("Bad encoding"),
+                tr("Could not read as {} text.\nYou could try another encoding").format(encoding),
```
aivarannamaa, line 251: Most of the translatable strings contain the old style formatting markers, so I suggest you leave `%s` here as well.

thonny/plugins/cpython_frontend/cp_front.py
```diff
@@ -241,7 +241,7 @@ def __init__(self, master):
             "You can activate an existing virtual environment also"
             + " via the right-click context menu in the file navagation"
             + " when selecting a virtual environment folder,"
-            + " or the 'pyveng.cfg' file inside."
```
aivarannamaa, line 244: Thanks for noticing this typo!

thonny/plugins/micropython/pip_gui.py
```diff
@@ -86,15 +86,15 @@ def _show_instructions_about_target(self):
...
-        self._append_info_text("Under the hood " + "\n", ("caption", "right"))
+        self._append_info_text(tr("Under the hood") + "\n", ("caption", "right"))
         self._append_info_text(
-            "This dialog uses `pipkin`, a new command line tool for managing "
-            "MicroPython and CircuitPython packages. ",
+            tr("This dialog uses `pipkin`, a new command line tool for managing "
+            "MicroPython and CircuitPython packages."),
             ("right",),
         )
         self._append_info_text("See ", ("right",))
```
aivarannamaa, line 95: I suppose you have omitted "See .... for more info" because of the piecewise placement of the whole sentence. You could still make it translatable by translating and splitting like this 

```python
trans = tr("See %s for more info")
prefix = trans.split("%s")[0].strip()
suffix = trans.split("%s")[1].strip()
```

thonny/plugins/micropython/pip_gui.py
```diff
@@ -86,15 +86,15 @@ def _show_instructions_about_target(self):
 
     def _show_extra_instructions(self):
         self._append_info_text("\n\n")
-        self._append_info_text("Under the hood " + "\n", ("caption", "right"))
+        self._append_info_text(tr("Under the hood") + "\n", ("caption", "right"))
         self._append_info_text(
-            "This dialog uses `pipkin`, a new command line tool for managing "
-            "MicroPython and CircuitPython packages. ",
+            tr("This dialog uses `pipkin`, a new command line tool for managing "
```
aivarannamaa, line 91: Please remove the word "new" from this text. It has to be removed at some point, but removing it later would invalidate all translations written so far.

thonny/plugins/micropython/uf2dialog.py
```diff
@@ -161,14 +161,14 @@ def _update_target_info(self):
...
+                text = tr("Unknown board")
                 label = "info"
 
         elif not self._target_combo.mapping:
-            text = "[no suitable targets detected]"
+            text = "[" + tr("no suitable targets detected") + "]"
             label = ""
         else:
-            text = f"[found {len(self._target_combo.mapping)} targets, please select one]"
+            text = "[" + 'found {} targets, please select one'.format(len(self._target_combo.mapping)) + "]"
```
aivarannamaa, line 171: did you forget `tr`?

thonny/plugins/micropython/uf2dialog.py
```diff
@@ -216,15 +216,15 @@ def _present_variants_for_target(self, target: TargetInfo) -> None:
...
+            enhanced_mapping = {"--- " + tr('MOST POPULAR ').ljust(100, '-'): {}}
             for variant in populars.values():
                 popular_variant = variant.copy()
                 # need different title to distinguish it from the same item in ALL VARIANTS
                 popular_title = self._create_variant_description(variant) + " "
                 popular_variant["title"] = popular_title
                 enhanced_mapping[popular_title] = popular_variant
 
-            enhanced_mapping["--- ALL VARIANTS " + "-" * 100] = {}
+            enhanced_mapping["--- " + tr('ALL VARIANTS ').ljust(100, '-')] = {}
```
aivarannamaa, line 227: better move the trailing space outside of the `tr` argument

aivarannamaa (changes requested):
Thank you!

Please see my comments. 

(In the future I recommend to make a proposal before undertaking a nontrivial i18n effort, because I've left some strings untranslatable on purpose, eg. because I expect them to change and don't want to add extra work for dozens of translators just by invalidating it soon after)

bletvaska:
ahoj

and thanks for the reply. initially my goal was to improve slovak translation, but then i found more strings, which are normally visible for the users, untranslatable, so i started digging in the code. meanwhile i've was digging again, but i decided to wait for your answer. so - thanks - i'll write my answers to your comments.

thonny/codeview.py
bletvaska: Yes, I saw it. My decision was based about my experiences. When I teach Python programming and we talk about string formatting in Python, this is something I am not showing to students, or at least I notice them about not to use this style of formatting. I've seen some articles mentioning about not using this style, and also in the official documentation is mentioned, that using such style can be tricky in some situations. Also, when I was translating Mu editor, they were not using "old style" but new one. Question can be about the mentioning the parameter inside of the brackets, so instead of translating string in the form:

```
"Could not read as {} text.\nYou could try another encoding"
```

will be much more "readable" and "explainable" to have string in format:

```
"Could not read as {encoding} text.\nYou could try another encoding"
```

But the final decision is up to you. So I can put it back to "old style".

thonny/plugins/micropython/pip_gui.py
bletvaska: My original idea was just about making "visible strings" translatable without any big effort ;) I can take a closer look (also about how tkinter works).

thonny/plugins/micropython/pip_gui.py
bletvaska: makes sense

thonny/plugins/micropython/uf2dialog.py
bletvaska: Hmm... It looks like...

thonny/plugins/micropython/uf2dialog.py
bletvaska: You are right. I was experimenting a bit to make it working. And I think, my decision was based on some strings, which also ends with trailing space. I'll fix it.

bletvaska:
Yes - I know about POEditor. My Slovak translation was imported in there.

aivarannamaa:
I now added your changes manually to current master (ab053975576d57eff72987b19c077ff5cf13e00a). Thanks for your work.

Issue #2680, opened 2023-03-02
todo view


kr-g:
when using todo view a line such as

`# todo text `

is shown in todo view

whereas a line such as 

`# text todo`

is not shown in the todo view

aivarannamaa:
Thanks!

PR #2681, opened 2023-03-02
Fix todo view

kr-g:
belongs to
https://github.com/thonny/thonny/issues/2680


kr-g:
changed regex to match the requirement

Issue #2650, opened 2023-02-10
Thonny on raspberrry pi running pop_os gives error when looking for connected pico


stormagedon92:
when i click the "Local Python 3 - /usr/bin/python3" i get this error message:

Traceback (most recent call last):
  File "/app/lib/python3.10/site-packages/tkinter/__init__.py", line 1705, in __call__
    return self.func(*args)
  File "/app/lib/python3.10/site-packages/thonny/workbench.py", line 865, in _post_backend_menu
    if uf2_device_is_present_in_bootloader_mode():
  File "/app/lib/python3.10/site-packages/thonny/plugins/micropython/uf2dialog.py", line 602, in uf2_device_is_present_in_bootloader_mode
    for vol in list_volumes(skip_letters=["A"]):
  File "/app/lib/python3.10/site-packages/thonny/misc_utils.py", line 101, in list_volumes
    mount_output = subprocess.check_output("mount").splitlines()
  File "/usr/lib/python3.10/subprocess.py", line 420, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.10/subprocess.py", line 501, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.10/subprocess.py", line 969, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.10/subprocess.py", line 1845, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'mount'

aivarannamaa:
Thank you for the report!
What do you get if you run command 'mount' in terminal?

stormagedon92:
Thanks for the quick reply, this is what I get:

sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime)
proc on /proc type proc (rw,nosuid,nodev,noexec,relatime)
udev on /dev type devtmpfs (rw,nosuid,relatime,size=3788624k,nr_inodes=947156,mode=755,inode64)
devpts on /dev/pts type devpts (rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000)
tmpfs on /run type tmpfs (rw,nosuid,nodev,noexec,relatime,size=794432k,mode=755,inode64)
/dev/mmcblk0p2 on / type ext4 (rw,relatime)
[...]

Issue #2426, opened 2022-09-04
Raspberry Pi Pico does not mount


ottlukas:
Internal Error

```
Traceback (most recent call last):
  File "/app/lib/python3.10/site-packages/tkinter/__init__.py", line 1705, in __call__
    return self.func(*args)
  File "/app/lib/python3.10/site-packages/tkinter/__init__.py", line 749, in callit
    func(*args)
  File "/app/lib/python3.10/site-packages/thonny/workdlg.py", line 175, in _keep_updating_ui
    self.update_ui()
  File "/app/lib/python3.10/site-packages/thonny/plugins/micropython/uf2dialog.py", line 92, in update_ui
    targets = self.find_targets()
  File "/app/lib/python3.10/site-packages/thonny/plugins/micropython/uf2dialog.py", line 121, in find_targets
    for vol in list_volumes(skip_letters=["A"])
  File "/app/lib/python3.10/site-packages/thonny/misc_utils.py", line 98, in list_volumes
    mount_output = subprocess.check_output("mount").splitlines()
  File "/usr/lib/python3.10/subprocess.py", line 420, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.10/subprocess.py", line 501, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.10/subprocess.py", line 969, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.10/subprocess.py", line 1845, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'mount'
```
Thonny Version: 4.0.0 (installed via Flathub)
Operating System: Linux Mint 21                   
Kernel: Linux 5.15.0-47-generic
Python 3.10.6 (64 bit)
Tk 8.6.12

```
Couldn't find the device automatically. 
Check the connection (making sure the device is not in bootloader mode) or choose
"Configure interpreter" in the interpreter menu (bottom-right corner of the window)
to select specific port or another interpreter.
```
Anyone has an idea how I can debug / fix this internal errror in Thonny to flash my Raspberry Pi Pico? - Thank you!

ottlukas:
Ok the issue was cable related see also: https://support.thepihut.com/hc/en-us/articles/360017234598-My-computer-won-t-detect-my-Raspberry-Pi-Pico

[...]

PR #2683, opened 2023-03-04
Use UDisks to find mounted filesystems on Linux

### LINKED ISSUE #2649: Could not find volume CIRCUITPY
I am unable to write to the device using the IDE. I am getting the below error when I press the run button:

```
PROBLEM IN THONNY'S BACK-END: Exception while handling 'write_file' (RuntimeError: Could not find volume CIRCUITPY).
See Thonny's backend.log for more info.
You may need to press "Stop/Restart" or hard-reset your CircuitPython device and try again.


Process ended with exit code 1.
```

Version of Thonny: 4.0.2
Version of CircuitPython: 8.0.0

jwillikers:
This should generally work better on Linux than using the mount command directly. UDisks is the utility that all major Linux desktop distributions use to handle mounting filesystems in userspace. UDisks is accessible through a D-Bus API on the system bus.

D-Bus is complex and requires adding a dependency on a Python library for handling asynchronous communication with the bus. I chose dbus-next since it is actively developed, modern, and doesn't require a glib main loop. Since Thonny is not a Gtk application, it makes more sense to use native Python async functionality instead.

Inside a Flatpak, it's necessary to use UDisks via the D-Bus API to mount and check mounted filesystems. The Flatpak sandbox prevents mounting otherwise and does not supply a mount command with the runtime.

Fixes flathub/org.thonny.Thonny#153, #2650, and #2426.
Probably fixes #2649.


aivarannamaa:
Thanks, and sorry for the late response!

It looks like `python-dbus-next` isn't too heavy and doesn't pull in other deps. In this case I'm happy to try this as the fallback when `mount` is not available (and possibly as the only option in Thonny 4.1 or 5.0).

jwillikers:
> Thanks, and sorry for the late response!
> 
> It looks like `python-dbus-next` isn't too heavy and doesn't pull in other deps. In this case I'm happy to try this as the fallback when `mount` is not available (and possibly as the only option in Thonny 4.1 or 5.0).

Great! I was thinking I would attempt to use Udisks before falling back to `mount` on Linux. So far, I've got a standalone `udisks.py` file in this MR that I can run locally and get the file path to an attached RP2040 board with MicroPython. I just need to find the best way to integrate this with the current filesystem-mounting logic and how to integrate the necessary async event loop. I'll also be adding more robust error handling. I'm hoping to find the time to finish this in the next couple of weeks. Let me know it you have any suggestions or advice. Thank you much!

jwillikers:
@aivarannamaa I've got this working now. I've tested it in the Flatpak and verified that flashing a Raspberry Pi Pico in the UF2 dialog works without issue.

CapeCrusader321:
Should we also consider adding support of xdg desktop portals ??

jwillikers:
> xdg desktop portals

I believe Tcl/Tk already uses the native file-chooser dialog when selecting a file, which is why it is possible to select files to open and save when the Flatpak sandbox is completely isolated. However, for the Flatpak I've allowed permissions to the user's home directory and the default locations for mounted media. This is so that paths to files are consistent with a person's directory layout on their machine and to allow access to mounted devices. Having Python files individually opened and mounted to arbitrary locations inside the Flatpak makes it untenable to develop anything in Thonny larger than a single file. Most development Flatpaks grant even greater permissions by default for the same reason. I don't think there is an XDG desktop portal which allows opening directories either, which would be the primary solution if the sandbox were to be restricted further.

I'm not sure which XDG desktop portals you have specifically in mind, but I've assumed so far you were referring to file access. Was there something more specific to this PR you had in mind or were you asking more generally?

CapeCrusader321:
> > Should we also consider adding support of xdg desktop portals ??
> 
> I believe Tcl/Tk already uses the native file-chooser dialog when selecting a file, which is why it is possible to select files to open and save when the Flatpak sandbox is completely isolated. However, for the Flatpak I've allowed permissions to the user's home directory and the default locations for mounted media. This is so that paths to files are consistent with a person's directory layout on their machine and to allow access to mounted devices. Having Python files individually opened and mounted to arbitrary locations inside the Flatpak makes it untenable to develop anything in Thonny larger than a single file. Most Flatpaks of developer tools grant even greater permissions by default for the same reason. I don't think there is an XDG desktop portal which allows opening directories either, which would be the primary requirement to allow the sandbox to be restricted further.
> 
> I'm not sure which XDG desktop portals you have specifically in mind, but I've assumed so far you were referring to file access. Was there something more specific to this PR you had in mind or were you asking more generally?

What i was saying is this although thonny uses xdg portals for file access in flatpaks due to the default mechanism of flatpak thats built around xdg portals ,by default thonny does not uses xdg portals since in snap sandbox it continues to use the tk one, this is even after forcing it to use portals, thonny always uses its internally built less interactive yet working tk dialogs to access files, so i was thinking if we somehow force thonny to use xdg portals as its default globally, this was a suggestion or maybe a general question...

CapeCrusader321:
> > > > Should we also consider adding support of xdg desktop portals ??
> > > 
> > > 
> > > I believe Tcl/Tk already uses the native file-chooser dialog when selecting a file, which is why it is possible to select files to open and save when the Flatpak sandbox is completely isolated. However, for the Flatpak I've allowed permissions to the user's home directory and the default locations for mounted media. This is so that paths to files are consistent with a person's directory layout on their machine and to allow access to mounted devices. Having Python files individually opened and mounted to arbitrary locations inside the Flatpak makes it untenable to develop anything in Thonny larger than a single file. Most Flatpaks of developer tools grant even greater permissions by default for the same reason. I don't think there is an XDG desktop portal which allows opening directories either, which would be the primary requirement to allow the sandbox to be restricted further.
> > > I'm not sure which XDG desktop portals you have specifically in mind, but I've assumed so far you were referring to file access. Was there something more specific to this PR you had in mind or were you asking more generally?
> > 
> > 
> > What i was saying is that although thonny uses xdg portals for file access in flatpaks due to the default mechanism of flatpak being built around xdg portals ,by default thonny does not uses xdg portals since in snap sandbox it continues to use the tk one, this is even after forcing it to use portals, thonny always uses its internally built less interactive yet working tk dialogs to access files, so i was thinking if we somehow force thonny to use xdg portals as its default globally, this was a suggestion or maybe a general question...
> 
> Ahh, I think that's an aside from this PR, but do you have `zenity` installed and on `PATH` in the snap? That's necessary to get the native GNOME file chooser dialog in Thonny.

I will try pushing new build with staging zenity, thanks for the tip..

aivarannamaa:
Thanks!

At first try I got a segfault crash on Ubuntu 22.04, when opening the back-end selector menu (which looks for volumes looking like MicroPython or CircuitPython UF2 installation volumes). I couldn't reproduce it later, though.

I'll leave it as is for now, but if I experience this crash again, I'll make `mount` to be preferred approach and use udisks as a fallback.

Issue #2684, opened 2023-03-04
restart of thonny after changing env var in thonny options 


kr-g:
when changing an env variable in thonny > tools > general it requires a restart of thonny to reflect the new env settings for running scripts with thonny

aivarannamaa:
Thanks for the suggestion, but I have different plans regarding specifying environment variables for the backend process.

PR #2685, opened 2023-03-04
use settings env directly

kr-g:
https://github.com/thonny/thonny/issues/2684


kr-g:
https://github.com/thonny/thonny/issues/2684

aivarannamaa:
Thanks for the suggestion!

These environment variables were meant mostly for the frontend process (and therefore would be inherited by the backend process as a side effect). 

I'm considering adding the possibility for specifying environment variables for current launch configuration as well.

kr-g:
> I'm considering adding the possibility for specifying environment variables for current launch configuration as well.

having a additional text area ?
if you want to specify that more in detail and share your ideas i could offers some help here.

kr-g:
i opened that issue because i continued to look at 
https://github.com/thonny/thonny/pull/2060

and debugged there again thonnys code and paramiko

PR #2690, opened 2023-03-07
Changer "nœud" to "étape" in French translation of debuger actions.

nicolasdespres:
"nœud" (means "node") refers to the AST node. It it is a bit hard to explain to beginners what is a "node" in this context.
"étape" (means "step") refers to "(computation) step" which I find much more understandable for beginners.


aivarannamaa:
Thanks for the suggestion, but please make these changes in https://poeditor.com/join/project/Gh188fdYH6, which is the master source for translations.

nicolasdespres:
I have made the changes. Thx.

PR #2691, opened 2023-03-09
Add plugin for new Pololu 3pi+ 2040 Robot.

pdg137:
This is to support the new [Pololu 3pi+ 2040 Robot](https://www.pololu.com/product/5001).  It's mostly compatible with the Pico, so I copied most of it from the Pico plugin.  The main difference - and reason for making this as a separate option - is that it disables `supports_remote_directories` and `supports_remote_files`. The 3pi+ is shipping with a FAT filesystem, so we expect users to have that mounted in their OS so they can edit files directly, and it's very confusing to have Thonny keep offering to save files "on the board" as an alternative.

I would appreciate advice on how to make the firmware downloads accessible, too.  The current version is listed here:

https://www.pololu.com/docs/0J86/5.1

but we could make a special page for Thonny to access or possibly do something with GitHub if that would be helpful.

### DIFF
--- a/thonny/plugins/pololu_3pi_2040_robot/__init__.py
+++ b/thonny/plugins/pololu_3pi_2040_robot/__init__.py
@@ -1,0 +1,70 @@
+from logging import getLogger
+from typing import Optional
+
+from thonny import get_workbench, ui_utils
+from thonny.plugins.micropython import add_micropython_backend
+from thonny.plugins.micropython.mp_common import RAW_PASTE_SUBMIT_MODE
+from thonny.plugins.micropython.mp_front import get_uart_adapter_vids_pids
+from thonny.plugins.micropython.uf2dialog import Uf2FlashingDialog
+from thonny.plugins.rp2040 import RP2040BackendConfigPage, RP2040BackendProxy
+from thonny.ui_utils import show_dialog
+
+logger = getLogger(__name__)
+
+class Pololu3pi2040RobotBackendProxy(RP2040BackendProxy):
+    @classmethod
+    def should_consider_unknown_devices(cls):
+        return False
+
+    @classmethod
+    def get_known_usb_vids_pids(cls):
+        return {(0x1FFB, 0x2043)}
+
+    @classmethod
+    def get_node_label(self):
+        return "Pololu 3pi+ 2040 Robot"
+
+    # The 3pi+ robot uses a locally-mounted FAT filesystem.
+    # Disable "remote" file access to prevent confusion.
+    def supports_remote_directories(self):
+        return False
+
+    def supports_remote_files(self):
+        return False
+
+class Pololu3pi2040RobotBackendConfigPage(RP2040BackendConfigPage):
+    pass
+
+def load_plugin():
+    add_micropython_backend(
+        "Pololu3pi2040Robot",
+        Pololu3pi2040RobotBackendProxy,
+        "MicroPython (Pololu 3pi+ 2040 Robot)",
+        Pololu3pi2040RobotBackendConfigPage,
+        bare_metal=True,
+        sort_key="32",
+        validate_time=False,
+        sync_time=True,
+        submit_mode=RAW_PASTE_SUBMIT_MODE,
+        write_block_size=64,
+    )
+
+    # Don't consider Pico in generic backends
+    # The main reason is to reduce the number of items in the backend switcher menu
+    import thonny.plugins.circuitpython
+    import thonny.plugins.esp
+    import thonny.plugins.micropython
+    import thonny.plugins.rp2040
+
+    thonny.plugins.circuitpython.cirpy_front.VIDS_PIDS_TO_AVOID.update(
+        Pololu3pi2040RobotBackendProxy.get_known_usb_vids_pids()
+    )
+    thonny.plugins.micropython.mp_front.VIDS_PIDS_TO_AVOID_IN_GENERIC_BACKEND.update(
+        Pololu3pi2040RobotBackendProxy.get_known_usb_vids_pids()
+    )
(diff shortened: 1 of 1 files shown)


aivarannamaa:
Thanks for the suggestion!

I'll start with the easier topic -- firmware downloads. It would be convenient if you created a GitHub project and made a GitHub release with the UF2 assets, just like Pimoroni does: https://github.com/pimoroni/pimoroni-pico/releases. I would then modify [the script I'm using for pulling MicroPython download links](https://github.com/thonny/thonny/blob/master/data/update_micropython_variants.py) to consult your project as well. GitHub releases are convenient, because I could re-use the code I'm using for collecting Pimoroni links and I would watch for your releases and get automatically notified when a new release needs to be collected.

Regarding the new back-end -- I'd rather not add specific models to this list (MicroPython Pico is an exception, because it was introduced before the more general RP2040 back-end). But you have the option to make it a 3rd party plug-in. I want to propose a more general alternative, though...

I understand you don't want the users to have to choose between saving to "This computer" and "MicroPython device". I was in similar situation when creating the CircuitPython back-end. As you probably know, CircuitPython also presents its filesystem via a mounted volume. I decided it's better if Thonny's UI uses uniform approach for both MicroPython and CircuitPython -- i.e. shows separate "Device" pane in the file browser and presents the same target choice when saving a file. So far CircuitPython users haven't complained about this. (Note that Thonny can also handle the case when CircuitPython's fs is configured to be mounted read-only in order to make the filesystem writable by CircuitPython code.)

I believe at least some users like the separate "This computer" and device panes in Thonny's filebrowser, even if the device pane duplicates part of the "This computer" pane. Therefore I'm proposing adding a checkbox on the configuration page of RP2040 and CircuitPython back-ends, which allows turning off the separate device part of the filebrowser and also the extra choice when saving a file for the first time. You could then instruct your users to check this box. Maybe some CircuitPython users would like this as well.

pdg137:
Thanks for the detailed suggestions! I'll see about getting a GitHub repository set up for firmware releases.

I also get what you're saying about not wanting individual backends for every board. So are you open to adding our VID/PID combination to match the "MicroPython (RP2040)" option and hide the other ones like ESP32?

I didn't realize that Thonny's CircuitPython plugin also offered the "This Computer / RP2040 device" split. Is there some value I'm not understanding in other access method when the files are already mounted via FAT? In any case, if you can do the checkbox to disable it, that would be great. Ideally it would default to the correct setting, for example by auto-detecting a FAT filesystem in Python or identifying the mass storage interface of the USB device. Or you could have something clickable on that pane to close it.

Please let me know if there's anything I can do to help!

aivarannamaa:
> Is there some value I'm not understanding in the other access method when the files are already mounted via FAT? 

Actually, Thonny, in most cases, does use the mounted volume for writing the files in CircuitPython. It first tries to write via REPL but if it gets read-only filesystem error, it falls back to volume (which is faster anyway). As I said above, this two-step approach means Thonny is able to write the files even when the mounted volume is configured to be read-only and the REPL file commands become usable.

> Please let me know if there's anything I can do to help!

I'll give you an idea to consider... CircuitPython (which also mounts its filesystem to the dev machine) has a file called boot_out.txt, which contains information about the firmware, board and recently also unique CPU ID. For example:

```
Adafruit CircuitPython 8.0.3 on 2023-02-23; Pimoroni Tiny 2040 (2MB) with rp2040
Board ID:pimoroni_tiny2040_2mb
UID:DF6018F2674D3437
```

The addition of the third line was requested by myself in order to be able to correlate the serial port Thonny has connected to and the mounted volume. This is important when there are several CP boards connected at a time -- in this case Thonny needs a reliable means for determining which volume to use for the board it has currently connected to. For this it is supposed to query the CPU ID (`binascii.hexlify(microcontroller.cpu.uid).upper().decode("ascii")`) and locate the volume which contains boot_out.txt  with this ID on third line.

My suggestion is that you implement similar information out-file in your firmware. This way, the users who don't hide MicroPython filebrowser for Pololu board, will be able to use the same approach (UI-wise) as with other MicroPython boards.

pdg137:
We made a releases page that I hope you can use:

https://github.com/pololu/micropython-build/releases/

Also I released [a plugin](https://pypi.org/project/thonny-pololu/), since that turned out to be really easy.  Please let me know if you see any problems with those.

Your suggestion about boot_out.txt sounds reasonable to me but it seems like a thing for the MicroPython maintainers to implement more generally. And I'm not sure that MicroPython has the read-only locking option that CircuitPython has.  If Thonny can associate the two file access options does that mean it won't have the two file panes or ask whether to load/save files on the device?

aivarannamaa:
> We made a releases page that I hope you can use: https://github.com/pololu/micropython-build/releases/

Great! I now subscribed to the releases of this project and already uploaded the information about the current uf2 to https://github.com/thonny/thonny/blob/master/data/micropython-variants-uf2.json so you should be now able to install this MicroPython variant via Thonny's UF2 installation dialog.

> If Thonny can associate the two file access options does that mean it won't have the two file panes or ask whether to load/save files on the device?

No, this would only mean that Thonny would be able to find the mount after it has received read-only error from the REPL-based file write attempt. If this doesn't happen with Pololu, then it doesn't matter here. The extra pane and selection dialog will be hidden only by the new configuration checkbox we talked about earlier.

pdg137:
Thanks for updating the uf2 list so quickly!

We have updated [our instructions](https://www.pololu.com/docs/0J86/5.3) to refer to Thonny and the thonny-pololu plugin, so I think we're in a good state now.  I'd be happy to discuss the checkbox idea more when you get to it, especially if you're interested in making the behavior depend on the device somehow.

PR #2699, opened 2023-03-11
Require saving before using fast debugger

michaellass:
In 0bb3b5fb a check was introduced to ensure that the program is saved before debugging. However, this check currently only catches attempts to run the nice debugger. The fast debugger also requires the file to be saved, so catch that one as well.


aivarannamaa:
Thank you!

PR #2703, opened 2023-03-13
Update to outline.py to skip multi-line strings

### LINKED ISSUE #2697: Outline pane ignores multi-line strings
The outline pane currently doesn't recognize multi-line strings. Consider the following situation:
```
usefullString = """This is a description of what this program does. It prints this string!
 class UselessClass is not even syntacitally a correct class and
 def uselessFunction(first, second):
     print("at least looks like a function definition")"""

class SemiUsefullClass:
    def __init__(self):
        print(usefullString)

uselessInstance = SemiUsefullClass()
```
In this case `uselessClass` and `uselessFunction` show up in the outline pane, while they should not (in my opinion).

*Possible solution:*

It would be easy to check for """ and skip the parsing until the next occurance of """.

However, the following approach would be a better solution (in my opinion):
`def _parse_source(self, source):` in `outline.py` should use a tokenizer instead of a simple single-line-matching of `r"[ \t]*(async[ \t]+)?(?P<type>(def|class){1})[ ]+(?P<name>[\w]+)`

P.S.: I suppose it might be easy to fix #385 along the way.

P.P.S.: Do you want me to try and fix this (can't promise that I won't break something, though 😉)? If so, which tokenizer should I use?

rauferd:
Makes the outline "parser" skip multi-line strings. It works for all, but the most obscure cases, and (mostly) fixes #2697

### DIFF
--- a/thonny/plugins/outline.py
+++ b/thonny/plugins/outline.py
@@ -80,8 +80,14 @@ class OutlineView(ttk.Frame):
         active_node = root_node
 
         lineno = 0
+        in_multiline_string = False
         for line in source.split("\n"):
             lineno += 1
+            if len(re.findall(r'\"\"\"', line)) % 2 == 1:      # an odd number of """s to toggle
+                in_multiline_string = not in_multiline_string  # multi-line string environment on/off
+                continue
+            if in_multiline_string:
+                continue
             m = re.match(r"[ ]*[\w]{1}", line)
             if m:
                 indent = len(m.group(0))


rauferd:
So far, I have not encountered problems with any "real life code" using this workaround in a couple of months.

From my point of view, this issue is fixed, the pull request will not be updated anymore and I abandoned the idea of programming a "real parser". Feel free to integrate this in next Thonny version.

If anyone is interested in programming a more sohpisticated parser, I am still willing to assist.

aivarannamaa:
Thank you! I'll include in Thonny 5.0

PR #2767, opened 2023-05-02
Close tab on middle click

mrexodia:
Simple quality of life feature that I missed a lot while using Thonny. Tested on Windows and macOS (with mouse and trackpad).


aivarannamaa:
Thanks! Seems to work nicely on macOS. On which platform did you test it?

PR #2768, opened 2023-05-02
Switch to the Tcl modernText library

### LINKED ISSUE #2639: Shift+Click selection starts from last mouse click instead of cursor position
If I hold down SHIFT and then left-click somewhere in my code, my expectation is that it should highlight all of the text between the current cursor position and where I've clicked.

Instead, it highlights the text between where I've just clicked, and the last place I clicked before that.  i.e. It thinks the current cursor position is still located where I last dropped it using the mouse, when in fact I could have moved the cursor using the keyboard since that time.

To reproduce:

1) Paste the following code into a Thonny editor window (e.g. main.py):
```
a = 1
b = 2
c = a + b
d = a + b + c
```
2) Left-click on the third line, just after the "a"
3) Press HOME, or press the left arrow 5 times, to move the cursor to the beginning of that line.  (Don't use the mouse for this step!)
4) Hold down your left SHIFT button, and while holding it, click the end of the very last line.
5) Observe that even though you clearly intended to select the full contents of the final two lines, the highlight only grabbed part of the third line.

Repeat the above steps in any other text editor (e.g. Notepad, a Chrome TEXTAREA, etc) to see how it _should_ function.

Note this was tested using Thonny 4.0.2 on Windows 7.

This is my first Thonny bug report, please be gentle and forgive me if I didn't fill out the correct template or whatever.

mrexodia:
Reference: https://wiki.tcl-lang.org/page/Modern+Bindings+for+the+Text+Widget

Closes #2639


aivarannamaa:
Thanks for this effort!

Unfortunately I can't promise I can review this soon (it need lot of attention) and I'm not even sure I dare to use this approach.

modernText.tcl seems to be rather old and I'm afraid we could lose some other fixes and updates if we switch.

Nevertheless, I'll take a look once I have more free time.

mrexodia:
What about merging it behind a checkbox? Most likely there is an updated version of `modernText`, although I think all of Tcl is actually ancient (2012 from what I can see) so I don’t think there will be any meaningful features added…

Moosems:
Certainly an interesting idea. It might be better to just make binds to the Text that override the default by using bind_class and therefore changing the default behavior.

Unrud:
This also fixes https://github.com/thonny/thonny/issues/2996.

mrexodia:
Any updates on this?

mrexodia:
I’m happy to rebase this or do additional tests etc. From my research the latest editor in Tcl is **much** older than modernText, but I’m happy to add a setting to revert to the default one.

aivarannamaa:
A rebase would be good!

mrexodia:
Alright, done!

thonny/workbench.py
```diff
@@ -181,6 +181,20 @@ def __init__(self, parsed_args: Dict[str, Any]) -> None:
 
         tk.Tk.__init__(self, className="Thonny")
         tk.Tk.report_callback_exception = self._on_tk_exception  # type: ignore
+
```
aivarannamaa, line 184: Could you please extract this into separate method?

thonny/tktextext.py
```diff
@@ -34,6 +34,15 @@ def __init__(self, master=None, cnf={}, read_only=False, suppress_events=False,
         self._original_mark = self._register_tk_proxy_function("mark", self.intercept_mark)
         self.bind("<Control-Tab>", self._redirect_ctrl_tab, False)
 
+        # Reference: https://wiki.tcl-lang.org/page/Modern+Bindings+for+the+Text+Widget
```
aivarannamaa, line 37: Is all of this required for each instance of Text?

aivarannamaa (changes requested):
Please check my comments

thonny/tktextext.py
mrexodia: If I remember correctly the `self.bindtags` is required every time a new text instance is created.

thonny/workbench.py
mrexodia: Done!

mrexodia:
I made the changes you requested, could you re-review @aivarannamaa?

aivarannamaa:
It looks like I botched the PR with my rebase somehow. I'll handle this.

aivarannamaa:
I now investigated https://wiki.tcl-lang.org/page/Modern+Bindings+for+the+Text+Widget more closely and noticed

> [KJN](https://wiki.tcl-lang.org/page/KJN): See [ntext](https://wiki.tcl-lang.org/page/ntext) for the completed version of this package.
> 
> * The tasks in the "To Do" list below have been completed (except writing the TIP; feedback from users would be helpful before the TIP is written).
> * The package has been renamed from "modernText" to "ntext" to avoid confusion with the Computer Modern fonts (my thanks to the person who pointed this out).

* https://core.tcl-lang.org/tklib/doc/trunk/embedded/md/tklib/files/modules/ntext/ntext.md
* https://core.tcl-lang.org/tklib/doc/trunk/embedded/md/tklib/files/modules/ntext/ntextBindings.md
* https://core.tcl-lang.org/tklib/doc/trunk/embedded/md/tklib/files/modules/ntext/ntextIndent.md
* https://core.tcl-lang.org/tklib/doc/trunk/embedded/md/tklib/files/modules/ntext/ntextWordBreak.md
* https://github.com/tcltk/tklib/tree/master/modules/ntext

Thanks for bringing my attention to this, but I'll close this PR, hoping to investigate ntext after Thonny 5.0 gets released.

Issue #2750, opened 2023-04-17
thonny 4.0.2: dialogs for opening/saving files don't work with zenity 3.91


befeleme:
Thonny in Fedora Linux 38 in Gnome environment, with zenity 3.91 doesn't open dialogs for opening or saving files:
Clicking on the buttons does nothing. Same when pressing the keyboard shortcuts.

This is logged when opening the file:
```
16:34:36.658 WARNING thonny.ui_utils: Zenity returned code 255 and stderr 'This option is not available. Please see --help for all possible usages.\n'
```
This, when saving:
```
16:35:58.465 INFO    thonny.editors: Save dialog returned None with typevariable ''
16:36:19.480 WARNING thonny.ui_utils: Zenity returned code 255 and stderr 'This option is not available. Please see --help for all possible usages.\n'
```
There is a known workaround (https://github.com/thonny/thonny/issues/1576#issuecomment-768346921) that works also in this case.
I decided to open the issue here as I'm not sure whether this is something to be fixed on Thonny's side or zenity's. Apparently, zenity 3.91 is still alpha (https://gitlab.gnome.org/GNOME/zenity/-/issues/54#note_1715285)

How to reproduce:
Run the Python Classroom Lab based on F38 where Thonny is preinstalled and try to open or save a file: https://labs.fedoraproject.org/prerelease 
I'll be happy to test the fix.

befeleme:
Minimally the buttons Open.. (ctrl+o), Save (ctrl+s ) and the respective options in the File menu: Open, Save, Save as, Save all files, Save copy. I have not tried everything, but New (ctrl+n), copying and pasting work correctly.

aivarannamaa:
@befeleme, @hroncok, could you please remove `--modal` and/or `--attach` parameters from the following location and try again: https://github.com/thonny/thonny/blob/be1536f930626275febc3d6a4d21327a2c0316f8/thonny/ui_utils.py#L2116 These options doesn't seem to have the expected effect anyway (at least on some systems).

If this doesn't help, then perhaps you spot another parameter in the same class which may be missing in the new zenity.

befeleme:
I did more digging in Fedora 39 with zenity 3.91:

- `--attach` results in error with the above mentioned message.
- `--modal` without `--attach` doesn't result in error (I don't know whether it's needed - without both `--modal` and `--attach` the dialog opens correctly)
-  `--name` and `--class` here:
[...]

PR #2786, opened 2023-05-16
Enable thonny to open and save files with zenity 3.91+

befeleme:
Based on the man pages of zenity and my own testing with zenity 3.91 and 3.92 on Fedora Linux 38, this is the minimum required to open and save files in the Gnome environment. Disclaimer: this was crafted using the trial & error method and I can't really tell whether this is a proper way to fix the issue long-term. Any guidance here will be much appreciated.

- Removed `--attach`, `--name` and `--class` options which are no longer recognized.

- Removed the deprecated option `--confirm-overwrite` (man zenity 3.92.0): --confirm-overwrite DEPRECATED; does nothing

Fixes: #2750

### DIFF
--- a/thonny/ui_utils.py
+++ b/thonny/ui_utils.py
@@ -2098,8 +2098,6 @@ class _ZenityDialogProvider:
     def asksaveasfilename(cls, **options):
         args = cls._convert_common_options("Save as", **options)
         args.append("--save")
-        if options.get("confirmoverwrite", True):
-            args.append("--confirm-overwrite")
 
         filename = cls._call(args)
         if not filename:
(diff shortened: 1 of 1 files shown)


aivarannamaa:
Thanks!

In the long run I have plans to implement a custom file picker for Linux, so your solution is good for now.

sanchosk:
Is it possible this problem is still there? I installed Thonny from debian repo (thonny: Thonny version: 4.0.1), but on saving, I get the error:
```
10:13:10.804 WARNING thonny.ui_utils: Zenity returned code 255 and stderr 'This option is not available. Please see --help for all possible usages.\n'
10:13:10.804 INFO    thonny.editors: Save dialog returned None with typevariable ''
```

aivarannamaa:
Debian doesn't have latest Thonny

PR #2822, opened 2023-06-17
adding Adafruit ESP32-S2 Reverse Feather to circuitpython devices.json

jposada202020:


aivarannamaa:
Thanks! Please note, that this json file is not used in newer Thonny versions. Now I'm scraping https://circuitpython.org/downloads for metadata which is used in new kind of flashing dialog.

jposada202020:
Thanks for the clarification.  Maybe need to update mine. I found this after was trying to flash it in the feather and the dialog box suggest that I made a PR.

Will take that into account.

Thank you very much for you work

PR #2890, opened 2023-08-05
Farsi translation update

CUMOON:
Translation and fixing


aivarannamaa:
Thanks, but did you also update the translations in the POEditor? Please see https://github.com/thonny/thonny/blob/master/docs/translate.md#short-story-for-translators

CUMOON:
Yes, I updated the translation in POEditor

CUMOON:
and also im sorry about that i didn't see this line

"Translations will be pulled from POEditor before each release, no need to create a PR."

PR #2939, opened 2023-09-23
Update languages.py

emryswaves:
Updated languages.py to reflect the correct way to say Romanian aka Română.


aivarannamaa:
Thanks!

PR #2985, opened 2023-10-31
run PowerShell instead of cmd on Windows

ChrisMayfield:
In Windows 10, PowerShell [replaced][1] Command Prompt (cmd) as the default shell. PowerShell has been [installed by default][2] for over 10 years (since Windows 7 SP1 and Windows Server 2008 R2 SP1). Opening a Terminal in VS Code runs PowerShell by default. However, opening a Terminal in Thonny runs cmd by default. PowerShell is better for classrooms with a mix of Windows, Mac, and Linux computers, because the same commands can be run on all three OS's.

This pull request makes three minor changes to terminal.py:
1. Adds `_get_windows_terminal_command()` to return either `"pwsh"` (if the latest version is installed) or `"powershell"` (the pre-installed version).
2. Modifies `run_in_terminal()` and `_run_in_terminal_in_windows()` to use PowerShell instead of cmd. This change removes the "weird quoting" and relies on `subprocess.Popen()` to correctly quote the arguments.
3. Removes the `title` parameter from `run_in_terminal()`, which could optionally be set for Windows terminals only, but is now obsolete. As far as I can tell, the title was always `None` anyway.

[1]: https://support.microsoft.com/en-us/windows/powershell-is-replacing-command-prompt-fdb690cf-876c-d866-2124-21b6fb29a45f
[2]: https://learn.microsoft.com/en-us/powershell/scripting/windows-powershell/install/installing-windows-powershell?view=powershell-7.3

### DIFF
--- a/thonny/terminal.py
+++ b/thonny/terminal.py
@@ -5,7 +5,7 @@ import subprocess
 import sys
 
 
-def run_in_terminal(cmd, cwd, env_overrides={}, keep_open=True, title=None):
+def run_in_terminal(cmd, cwd, env_overrides={}, keep_open=True):
     from thonny.running import get_environment_with_overrides
 
     env = get_environment_with_overrides(env_overrides)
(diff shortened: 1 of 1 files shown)


aivarannamaa:
Thank you and sorry for the late reply!

I'm a bit hesitant in ditching cmd altogether, but let's make a try. If people start complaining, I'll make the shell configurable.

PR #3019, opened 2023-11-25
Update translation

Addemkrt:
added missing/wrong translation.


aivarannamaa:
Thanks, but please see https://github.com/thonny/thonny/blob/master/docs/translate.md

As POEditor is the main source for translations, you should add your changes there, otherwise they would be overwritten next time I'll import the translations to Thonny.

PR #3025, opened 2023-11-29
Superfluous comma

bmcage:
Error on linux due to extra property in json expected due to ,

### DIFF
--- a/data/micropython-variants-daplink.json
+++ b/data/micropython-variants-daplink.json
@@ -43,7 +43,7 @@
         "downloads": [
             {
                 "version": "2.1.2",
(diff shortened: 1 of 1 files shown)


bmcage:
When I install a tag instead, eg v1.4.1 the error downloading the json remains, as /master/ is hardcoded in plugins/__init__.py

A good improvement extra would be to download from the tag, instead of master, so something like:

    def get_variants_url(self) -> str:
        return f"https://raw.githubusercontent.com/thonny/thonny/v4.1.4/data/{self.firmware_name.lower()}-variants-daplink.json"

aivarannamaa:
Thanks!

> A good improvement extra would be to download from the tag

Do you mean Thonny should download from a tag? This would beat the purpose of keeping this json separate from Thonny's file. When it is downloading from a branch, I can publish new json without making a new release.

Issue #3002, opened 2023-11-15
Update pattern for Pololu 3pi+ and add new Zumo 2040 robot


pdg137:
Hello, regarding the Pololu firmware variant entry here:

https://github.com/thonny/thonny/blob/master/data/update_micropython_variants.py#L237

Could you update it to match our latest release?

https://github.com/pololu/micropython-build/releases/

There should now be two entries, like this:

```python
pololu_variants = [
    {
        "_id": "pololu-3pi-2040-robot",
        "model": "3pi+ 2040 Robot",
    },
    {
        "_id": "pololu-zumo-2040-robot",
        "model": "Zumo 2040 Robot",
    },
]
```

and the version string needs to be updated to work with multiple MicroPython versions (right now it looks like it's hard-coded to 1.20.0).  I think you could just extend the version-matching part so that the version looks like `v1.22.0-preview-231113` if that works for you.

aivarannamaa:
I decided it's easier if in the update script I simply list the desired Pololu items as they should appear in the data file. Done now.

pdg137:
Thanks!

Is there something else we could do to make our releases work better with Thonny? I think the only reason we're publishing them on GitHub is to support these automatic updates. It would be nice to have new ones show up automatically and maybe also to see more of the old ones, too.

aivarannamaa:
You could also make a pull request for the Pololu section of the update script: https://github.com/thonny/thonny/blob/b3c3b8d2f63558eeed8ed378cf5e86b03508741a/data/update_micropython_variants.py#L224

This way I can put it online faster.

pdg137:
Okay! I'll do that if necessary, but please let me know if there's something we can do to fix the automatic updates. I thought you'd just be able to use a pattern like

```python
"_download_url_pattern": "https://github.com/pololu/micropython-build/releases/download/{version}/micropython-{id}-v{version}.uf2",
```

and then it would be easier for you since you wouldn't ever have to manually update anything.

pdg137:
Oh! Looking at the URL, maybe I can see the problem. Would it be easier for you if we named the releases like `v1.22.0-preview-231113` instead of just `231113`?

aivarannamaa:
Would the `preview` make sense for next releases as well? I proposed PR-s because then you wouldn't need to create GitHub releases but could upload each release wherever and name it whatever.

pdg137:
[...]

PR #3078, opened 2024-01-19
Update Pololu links

pdg137:
This adds the latest 240117 MicroPython v1.22.1 firmware releases for the 3pi+ and Zumo robots as well some older releases from:

https://github.com/pololu/micropython-build/releases/

See https://github.com/thonny/thonny/issues/3002 for earlier discussion.


aivarannamaa:
Thanks!

aivarannamaa:
I now updated the links file as well, so your new MicroPython variants should be available for users.

PR #3098, opened 2024-01-31
Update appstream metadata

jwillikers:
Add missing `developer_name` attribute to the appstream metadata which is now required.
See the build failure here for reference: https://buildbot.flathub.org/#/builders/28/builds/8761


aivarannamaa:
Thanks!

Issue #3110, opened 2024-02-11
Thonny fails to boot when /usr/bin/activate exists (conda)


scottfp:
Thonny fails to boot when /usr/bin/activate exists (conda)
---
May be a continuation of https://github.com/thonny/thonny/issues/2516.

## Description

After Thonny checks for `THONNY_USER_DIR`, and portable mode, it checks if a virtual environment is executable without actually checking if it's in a virtual environment. This makes Thonny set the user dir to `/usr/.thonny` on my system, which is not writeable by normal users. 

On the latest version (v4.1.4), there is a [function for this](https://github.com/thonny/thonny/blob/v4.1.4/thonny/common.py#L802-L808) but it's not actually being run. My local version (v4.0.2) does not have this function at all. Setting `THONNY_USER_DIR` dodged this issue, pinpointing it to `_compute_thonny_user_dir()` pretty quickly.

## Proposed solution

I tested the following approaches successfully by modifying site-packages locally, and I plan to prepare a PR when time permits. I personally prefer Option 1, as it's a more precise fix to this specific issue.

**Option 1 - Call `running_in_virtual_environment()` from `__init__.py` ([Ref](https://github.com/thonny/thonny/blob/v4.1.4/thonny/__init__.py#L87C10-L87C31))**

```diff
def _compute_thonny_user_dir():
    if os.environ.get("THONNY_USER_DIR", ""):
        return os.path.expanduser(os.environ["THONNY_USER_DIR"])
    elif is_portable():
        if sys.platform == "win32":
            root_dir = os.path.dirname(sys.executable)
        elif sys.platform == "darwin":
            root_dir = os.path.join(
                os.path.dirname(sys.executable), "..", "..", "..", "..", "..", ".."
            )
        else:
            root_dir = os.path.join(os.path.dirname(sys.executable), "..")
        return os.path.normpath(os.path.abspath(os.path.join(root_dir, "user_data")))
-   elif is_virtual_executable(sys.executable) and not is_private_python(sys.executable):
+   elif (
+       running_in_virtual_environment()
+       and is_virtual_executable(sys.executable)
+       and not is_private_python(sys.executable)
+   ):
        return os.path.join(sys.prefix, ".thonny")
    elif sys.platform == "win32":
        return os.path.join(_get_roaming_appdata_dir(), "Thonny")
    elif sys.platform == "darwin":
        return os.path.expanduser("~/Library/Thonny")
    else:
[...]

PR #3112, opened 2024-02-12
thonny/thonny#3110 - check if in virtual environment on boot

scottfp:
Fixes https://github.com/thonny/thonny/issues/3110

---
**How to test:**

1. Make sure you are on Linux and are using system python
2. Pull latest source code and open it
    ```
    git clone https://github.com/thonny/thonny thonny-issue-3110
    cd thonny-issue-3110
    ```
3. Create `/usr/bin/activate`
    ```
    sudo touch /usr/bin/activate
    ```
4. Run thonny to confirm the failure
    ```
    python -m thonny
    ```
    Output:
    ```
    PermissionError: [Errno 13] Permission denied: '/usr/.thonny'
    ```
5. Merge this patch locally
    ```
    # Add patch remote and fetch
    git remote add patch https://github.com/scottfp/thonny
    git fetch patch
    
    # Double check patch is not nefarious
    git diff patch/3110-check-if-in-venv

    # Merge patch
    git merge patch/3110-check-if-in-venv
    ```
6. Run thonny again, it should boot successfully
    ```
    python -m thonny
    ```
7. Clean up (be careful!)
    ```
    cd ..
    rm -r thonny
    sudo rm /usr/bin/activate
    ```

---

I ran `black` and `isort` on this file. Feedback is welcome :)

### DIFF
--- a/thonny/__init__.py
+++ b/thonny/__init__.py
@@ -10,7 +10,7 @@ sys.path.insert(
     os.path.normpath(os.path.join(os.path.dirname(__file__), "vendored_libs")),
 )
 
-from thonny.common import is_private_python, is_virtual_executable
+from thonny.common import is_private_python, is_virtual_executable, running_in_virtual_environment
 
 _last_module_count = 0
 _last_modules = set()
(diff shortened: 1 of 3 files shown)


aivarannamaa:
Thanks a lot for thorough explanation in #3110 and for this PR! Also, sorry for late reply!

The core of the issue is that `is_virtual_executable` was not correct. It turned out I don't need this method at all, as `running_in_virtual_environment` is a better substitute for all cases. I now pushed a commit which gets rid of `is_virtual_executable`.

Issue #3140, opened 2024-03-16
flatpak: Issue installing packages to virtual environment


jwillikers:
When I was responding to issue flathub/org.thonny.Thonny#259, I wanted to recommend installing packages via a virtual environment. However, when I attempted to do so in the Flatpak, through `Manage packages` I hit this error.

```
PROBLEM IN THONNY'S BACK-END: Exception while handling 'get_active_distributions' (ModuleNotFoundError: No module named 'pkg_resources').
See Thonny's backend.log for more info.

Process ended with exit code 1.
```

I also hit an error creating the initial virtual environment, but the environment was still created and interpreter opened in the virtual environment.
The error was related to upgrading `pip`.

```
Error: Command '['/run/user/1000/doc/307236b2/testing123/bin/python3', '-m', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1.
Error: process returned with code 1
```

Running that command inside the Flatpak manually has the following output:

```
$ flatpak enter org.thonny.Thonny bash
bash: /.cargo/env: No such file or directory
bash-5.2$ ls
0001-Update-appstream-metadata-3098.patch  bcrypt-cargo-sources.json  bundled-python-modules.json      flathub.json	     org.thonny.Thonny.yaml  thonny
README.md				  build-dir		     cryptography-cargo-sources.json  flatpak-builder-tools  repo		     thonny-sources.yaml
bash-5.2$ ls /run/user/1000/doc/fa19fe51/n-thonny-venv/bin/python
python      python3     python3.11  
bash-5.2$ ls /run/user/1000/doc/fa19fe51/n-thonny-venv/bin/python3
python3     python3.11  
bash-5.2$ /run/user/1000/doc/fa19fe51/n-thonny-venv/bin/python3 -m ensurepip --upgrade --default-pip
Looking in links: /tmp/tmpi9r0zfya
Processing /tmp/tmpi9r0zfya/setuptools-65.5.0-py3-none-any.whl
Processing /tmp/tmpi9r0zfya/pip-24.0-py3-none-any.whl
Installing collected packages: setuptools, pip
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
thonny 4.1.4 requires wheel, which is not installed.
Successfully installed pip-24.0 setuptools-65.5.0
```

I'm planning on investigating more but wanted to get an issue opened sooner rather than later.

jwillikers:
[...]

PR #3141, opened 2024-03-16
Create venv and install pip in venv as separate steps

jwillikers:
This fixes venv creation in the Flatpak for some reason. See issue #3140.

### DIFF
--- a/thonny/plugins/cpython_frontend/cp_front.py
+++ b/thonny/plugins/cpython_frontend/cp_front.py
@@ -21,7 +21,7 @@ from thonny.common import (
     running_in_virtual_environment,
 )
 from thonny.languages import tr
-from thonny.misc_utils import running_on_mac_os, running_on_windows
+from thonny.misc_utils import inside_flatpak, running_on_mac_os, running_on_windows
 from thonny.plugins.backend_config_page import (
     BackendDetailsConfigPage,
     TabbedBackendDetailsConfigurationPage,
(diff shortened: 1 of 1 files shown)


thonny/plugins/cpython_frontend/cp_front.py
```diff
@@ -356,6 +356,18 @@ def _create_venv(self, event=None):
             exe_path = os.path.join(path, "bin", "python3")
 
         if os.path.exists(exe_path):
+            proc = subprocess.Popen(
+                [exe_path, "-m", "ensurepip", "--default-pip", "--upgrade"],
```
aivarannamaa, line 360: > It appears that Python is installed in the Freedesktop SDK using the --without-ensurepip conf option as seen [here](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/master/elements/components/python3.bst?ref_type=heads#L20). 

How come `ensurepip` works in this case?

aivarannamaa: In this case I'd prefer using the work-around only for flatpak installations and leave the current procedure as is for other cases.

jwillikers: Yeah, that sounds good to me. I wouldn't be surprised if this workaround becomes unnecessary with next release of the Freedesktop Flatpak SDK. I'll add a check that only does this for the Flatpak.

aivarannamaa:
Thanks a lot!

PR #3172, opened 2024-04-26
thonny/backend.py: Fix SSH

leon-anavi:
This fixes the following issue when using Remote Python3 (SSH) interpreter in Thonny:

Can't connect to '192.168.1.101' with user 'root': No existing session Re-check your host, authentication method, password or keys.

Process ended with exit code 1.

The fix sets to False look_for_keys and allow_agent in method connect of class Client from Paramiko when using password as an authentication methon. By defaut they are set to True.

References:
https://www.linuxfixes.com/2022/01/solved-python-paramikosshexceptionsshex.html https://github.com/ktbyers/netmiko/issues/1034


aivarannamaa:
Thank you for the PR and sorry for the late reply!

Are you sure this solution doesn't break authenticating with a private key? Perhaps setting `look_for_keys=False` and `allow_agent=False` should be done only if configuration parameter `"SshCPython.auth_method" == "password"`?

PR #3176, opened 2024-04-27
--class, --name, --attach args no working on ubuntu 24.04 noble numbat LTS, remove extra zenity args

Giuseppe7887:
looks like zenity not working with --class and --name args and --attach is deprecated in version 4.0.1 on the latest ubuntu 24.04 noble numbat, add version control before to open file selection and remove --attach option


aivarannamaa:
It looks like you applied your changes to an older version of Thonny. In the master branch these changes have been applied already in #2786

PR #3221, opened 2024-06-16
Make file_type to recognize more python files

### LINKED ISSUE #3220: Highlightning not only on file extension
I'm trying out Thonny, and so far I really really like it. But, one issue I have is with the syntax highlightning…

It seems it's based off the file extension, and not by the actual language.

_Example_:
- Open a new file (“&lt;untitled&gt;”), and write some py code. All colored
- Save as `foo`. Color is gone
- Rename it to `foo.py`, and the color is back

Is there any way for you to change this? Either so I can set filetype manually (like many other editors have that option if the program doesn't pick it up automatically). Or if Thonny could start recognizing the files by the shebang (ie. whatever is on the 1st line, like: `#!/usr/bin/env python, <!doctype html>, etc`).

Or add it a fallback option - like, when not recognized, treat it like a python file.

Best regards,
 · Eric

―

// btw: In the wiki: [SeparateInstall#linux](https://github.com/thonny/thonny/wiki/SeparateInstall#linux)
// The line: “3. `. thonny_venv/activate`”, …should be: `. thonny_venv/bin/activate`

iefdev:
The `file_type` is just based on file extension, and maybe not by
 its actual file type. This will also allow backup files, as in: 
`*.py.bak, *.py.orig, etc`. It will also check the first line for the env in the “shebang”.

- `get_dots`: get the number of dots to use in `ext` (ie. `foobar.py.orig` will also pick up `py`)
- `is_python_file`:  returns True if `python` is found

Closes #3220


iefdev:
So, I managed to find a good solution to my issue (#3220) . First I tried  a different approach, with a second var: `PYTHON_FILES_PATTERN = set()` and matching a few expressions. But, I think this is better and a more _generic_ solution.

Hope you'll like the changes. :+1: 

Best regards,
 · Eric

thonny/editors.py
```diff
@@ -92,8 +92,15 @@ def update_file_type(self):
...
             self._code_view.set_file_type(None)
         else:
-            ext = self._filename.split(".")[-1].lower()
-            if ext in PYTHON_EXTENSIONS:
+            # Include backup files: *.py.bak, *.py.orig, etc
+            get_dots = len(re.findall(r'\.', self._filename))
+            ext = self._filename.split(".")[-get_dots].lower()
+
+            # Extract the 1st line to get the env
+            with open(self._filename, 'r') as shebang:
```
aivarannamaa, line 100: You should consult the editor content, not the file, as the file may be located in another machine.

thonny/editors.py
```diff
@@ -92,8 +92,15 @@ def update_file_type(self):
...
+                is_python_file = bool('python' in shebang.__next__().strip())
```
aivarannamaa, line 101: `bool` is not required, as `in` operator already returns a `bool`.

I suggest `has_python_shebang` as the variable name, because the `ext in PYTHON_EXTENSIONS or is_python_file` part sounds like "an animal is zebra if it has black and white stripes or it is zebra" ;)

thonny/editors.py
```diff
@@ -92,8 +92,15 @@ def update_file_type(self):
         if self._filename is None:
             self._code_view.set_file_type(None)
         else:
-            ext = self._filename.split(".")[-1].lower()
-            if ext in PYTHON_EXTENSIONS:
+            # Include backup files: *.py.bak, *.py.orig, etc
+            get_dots = len(re.findall(r'\.', self._filename))
+            ext = self._filename.split(".")[-get_dots].lower()
```
aivarannamaa, line 97: If the filename is "blah.blah.py", the extension would be "blah"? This doesn't look right.

aivarannamaa (changes requested):
Thanks for the PR!

Please check my comments.

thonny/editors.py
aivarannamaa: Please also consider this file named "my-favorite-programs":

```
python
cat
less
ssh
```

Would you want to color it as Python code?

iefdev:
Thanks for the comments/review. 👍 All good valid points. I'll work through these changes in the coming days. I might have to get back to you on the “editor content”, if I don't fig that out. :) I used `self._filename` since it was what in use within the class.

 · Eric

thonny/editors.py
iefdev: I've fixed all the other things, but this one… Well, there's been a lot of trial'n'error. :) I don't now, but I think I'm on to it. Just can't get the output/content. If you could provide some help/tips, I would _really_ appreciate that.

So, after trying a lot of different things - printing out a lot to terminal. With `self.notebook.get_editor(self.get_filename(), True)` I can get the editor, and I can get it to print out the filename, when I open a file.

```py
e = self.notebook.get_editor(self.get_filename(), True)

# e have: _name, _filename and _code_view
e._name                 # !editor2  
e._filename             # /path/2/my/file.py
e._code_view.text       # .!codeview2.!editorcodeviewtext
```

`e._code_view.text` have `get`, so I thought just grabbing the first line would be great, but it's empty.

```py
e._code_view.text.get('1.0', '1.end')    # empty
e._code_view.get_content()               # empty
```

Ideally, one would just want to use something like: `.text.get('1.0', '1.end').find('python')`. But, I'm not _too_ familiar with tk, so I feel like I'm a bit in the dark now - trying and guessing.

Any ideas? Or maybe I'm looking at the wrong place/function(s)?

Issue #3186, opened 2024-05-04
delete_word_left and delete_word_right unexpected behaviour


wohltat:
When deleting a word with `delete_word_left` (ctrl+backspace) or `delete_word_right` (ctrl+delete)  the behavior is not like in other editor and in my opinion faulty.

* When deleting back over a linebreak with multiple lines, it deletes all the whitespace + linebreak + the previous word. In sublime text for example this would be 3 steps.
* Strange behavior around brackets
  * It sometimes stops before the last bracket. When there is a word in the bracket like `[foo]`
  * deletion of adjacent brackets is also strange. For example deleting back in `foo bar []()` gives `foo ` even though there is a space in between.
  * When there are brackets as first characters in the console command line, then they cannot be deleted at all

wohltat:
This was bothering me quite a lot. So i had to do something.

Here is an algorithm where i  tried to copy the behavior of sublime text:
As far as i tested it, it is identical.

Basic idea is that there are 3 character groups. As long as the next character is of the same group the deletion continues. One preceding white-space is skipped so that it removed together with the word.

```python
import tkinter as tk

example_text = """

    start_index = text_widget.search(r'\w*\W?', cursor_index, backwards=True, regexp=True, stopindex='1.0')

    # If start_index is at the beginning of the text, move it to the beginning of the line
    if start_index == '':
        start_index = '1.0'
    else:
        start_index = text_widget.index(start_index + "+1c")

    # Delete text between start of the word and cursor position
    text_widget.delete(start_index, cursor_index)

# Example usage:
def on_ctrl_backspace(event):
    delete_word_left(text_widget)
    asdasdf
        aaaaa\tasdf\t \tfsdf\t\tsdksd\t     
    .)(=.,)[]$&/()''""....{} asdf§sadf
"""

def char_group(c):
    # group 1 = whitespaces, group 2 = symbols, group 3 = alphanum
    group2_chars = '!"#$%&\'()*+,-./:;<=>?@[\\]^`{|}~'
    if c.isspace(): return 1
    if c in group2_chars: return 2
    return 3

def delete_word_left(text_widget):
    # Get current cursor position
    cursor_index = text_widget.index(tk.INSERT)
    line_start_index = text_widget.index(cursor_index + " linestart")

    # get the current line and search from the end
    line = text_widget.get(line_start_index, tk.INSERT)
    cur_index = len(line)-1

    # step at least one whitespace at the beginning
    if line and line[cur_index].isspace():	
[...]

PR #3229, opened 2024-06-22
Improved delete word left/right behavior, that is ctrl+backspace/delete

wohltat:
Sublime text like behavior of ctrl+backspace/delete

https://github.com/thonny/thonny/issues/3186


aivarannamaa:
Thanks for the PR and sorry for the late reply!

I'll review this soon and probably include in Thonny 5.0

PR #3411, opened 2024-12-19
Update __init__.py

### LINKED ISSUE #3410: Thonny 4.1.4 (and other versions) installation problems error with windows user name create  app folder with user name
### Discussed in https://github.com/thonny/thonny/discussions/3249

<div type='discussions-op-text'>

<sup>Originally posted by **BitHacker49** July 11, 2024</sup>
Hi, I'm new to Thonny and am experiencing unsurmountable problems with installing it on a Windows 11 computer.

The installation process pretends to work smoothly, proudly closing with "Great success...!", but the result is a mess. As soon as I try to start Thonny, the error message below is its response. This happened to other versions (I think 3.1.2 and 3.2.7) before. When I deinstall Thonny, tons of key entries containing "Thonny" are left in the Windows registry. I managed to delete them manually but I hate to fiddle around with the registry, knowing close to nothing about it.

Any idea what's wrong with my computer?

image
</div>

t2solve:
try to fix #3410
will replace all weird character for username  on windows os via function remove_weird_characters
during create  AppData folders on __init__.py call

### DIFF
--- a/thonny/__init__.py
+++ b/thonny/__init__.py
@@ -2,6 +2,7 @@ import logging
 import os.path
 import sys
 import time
+import re
 from logging import getLogger
 from typing import TYPE_CHECKING, List, Optional, cast
 
@@ -51,6 +52,9 @@ _last_module_count = 0
 _last_modules = set()
 _last_time = time.time()
 
+def remove_weird_characters(input_string: str) -> str:
+    cleaned_string = re.sub(r'[^a-z]', '', input_string)
+    return cleaned_string
 
 def report_time(label: str) -> None:
     """
@@ -127,6 +131,7 @@ def get_ipc_file_path():
     import getpass
 
     username = getpass.getuser()
+    username = remove_weird_characters(username.lower())
 
     ipc_dir = os.path.join(base_dir, "thonny-%s" % username)
     os.makedirs(ipc_dir, exist_ok=True)


aivarannamaa:
Thanks!

Can you tell me, what do you get with `repr(getpass.getuser())` ?

t2solve:
for shure on my windows os 
i do get a really weird string
from 
```python 
repr(getpass.getuser())
``` 
the output is:
```  
'/.../XX=YY/CC=000/LIST=USER/USER'
``` 
my windows is integrated in a broader windows AD setup, somehow, i could not be more specific
and i could not influence what is set here....

but by the way, is \_\_init\_\_.py
```python 
thonny + 'anotherUserSpecificName'
#in line 121
ipc_dir = os.path.join(base_dir, "thonny-%s" % username)
``` 
already 'doubled user specific'  because
you want to create a folder inside C:\USER\USERNAME\AppData\Local\
so :
```
C:\USER\USERNAME\AppData\Local\thonny
```
is already user specific or ?
and 
```
C:\USER\USERNAME\AppData\Local\thonny-USERNAME
```
is then doubled or ? 
So an easy fix  could be 
```python 
ipc_dir = os.path.join(base_dir, "thonny-%s" % username)
```
to
```python 
ipc_dir = os.path.join(base_dir, "thonny")
```

aivarannamaa:
This part of code creates or selects a folder which is used for a lockfile, which is used for single instance behavior, so that when you double click a py-file in Windows Explorer, it gets opened in the same Thonny instance which is already open, instead of opening new Thonny window.

This lockfile used to be in regular Thonny conf directory, but I received an problem report about a school situation where somehow different users using different computers were using the same home directory (probably in a network share). That's why I added the username part to the path. I don't remember the details, but this trick solved that case.

So I'll keep this design, even if the username gets doubled in the path in most/normal cases.

But I like your proposal, and I'll merge it. Thanks!

PR #3563, opened 2025-03-03
Fix: Block commenting behavior compliant with PEP 8

### LINKED ISSUE #3404: Follow PEP 8 guideline for block comment indentation
Currently the comment toggle behavior is as follow:
```
def some_function():
#     some_variable = None
    pass
```

While it is _functional_, the resulting code isn't easy to look at, python learner will also assume that that is the way of doing comments in python. I suggest the commenting behavior to adhere to the PEP 8 guideline.

PEP 8 explicitly says that a block comment should be indented to the same level as the commented code:
```
def some_function():
    # some_variable = None
    pass
```

Reference: https://peps.python.org/pep-0008/#block-comments

WhitespaceQ:
Trying to fix #3404: Ensure commenting behavior adheres to PEP 8 guidelines

This pull request addresses the previously non-compliant commenting behavior in our codebase. The comments have been updated to meet PEP 8 standards. Both space and tab indentation have been tested and work correctly.

Please review the changes and let me know if they do not conform to any internal code guidelines. I am happy to make further adjustments to comply with the standards.

Thank you!


aivarannamaa:
Thanks for the PR and sorry for the late reply!

Unfortunately it is hard to see what is the real change, because you have changed string formatting in several places. Or does it matter for getting the desired change?

WhitespaceQ:
Functional commit is on https://github.com/thonny/thonny/pull/3563/commits/d007c5ab1ec986250da38e97345d05e82f20b572 the other commits are for style/readability only

WhitespaceQ:
I'm happy for another round of review to comply with your codebase quality, unfortunately serveral temporary variable (prefixed with `_` as per Python guidelines) is currently needed. Removing this is theoretically possible but I need to touch more codes than what is needed.

aivarannamaa:
Thank you and sorry for the late reply!

PR #3578, opened 2025-03-22
Remove unused and breaking compatibility imports

konflic:
Have no idea why, but found some unused imports that does not affect anything.

At leat one of them `from turtledemo.penrose import start` won't allow to start project on Python 3.12.3 on ubuntu 24.04


aivarannamaa:
Thank you!

Issue #3547, opened 2025-02-18
Keyboard Shortcuts Not Recognized When Using Hebrew Keyboard Layout


annonimono:
### **Keyboard Shortcuts Not Recognized When Using Hebrew Keyboard Layout**

**Description:**
When using the Hebrew keyboard layout in Thonny, keyboard shortcuts such as **Copy (Ctrl+C)**, **Cut (Ctrl+X)**, and **Paste (Ctrl+V)** are not recognized or do not work as expected.

**Steps to reproduce:**
1. Switch the keyboard layout to Hebrew.
2. Try using common shortcuts like **Ctrl+C**, **Ctrl+X**, and **Ctrl+V**.
3. Notice that these shortcuts are not functioning.

**Expected behavior:**
The **Copy**, **Cut**, and **Paste** shortcuts should work regardless of the active keyboard layout.

**Actual behavior:**
Shortcuts do not work when the Hebrew keyboard layout is active.

**Additional information:**
This issue occurs only when the Hebrew layout is in use. Switching to the default English layout restores the functionality of these shortcuts.

**Request:**
Please provide a list of all the available keyboard shortcuts for Thonny.

**Note:**
I am not very fluent in English, and I used GPT to help write this message.

aivarannamaa:
Thanks for the report! I'll investigate this later.

niknal357:
This is generally a wider issue (which affects users of non-QWERTY layouts). The correct fix is to read scancodes rather than keycodes for keybinds.

Issue #422, opened 2018-04-20
Edit keyboard shortcuts (Ctrl+C, Ctrl+V, etc.) should work with any language keyboard layout


aivarannamaa:
**[Original report](https://bitbucket.org/bitbucket-issue-migration\thonny-issues.zip/issue/422) by Anonymous.**

----------------------------------------

If the keyboard language is left to some non-latin (Greek, in my case) then Ctlr+C, Ctrl+A, Ctrl+V, etc. don't work as edit shortcuts. The user has to switch back to English for the shortcuts to work.

This is quite a nuisance in IDLE and I thought it would not exist in Thonny (Windows 7 OS). I type some code and switch to Greek in order to type some string. Then I select some text, press Ctrl+C, followed by Ctrl+V somewhere else, only to find out that nothing was copied to clipboard, because the Ctrl+C and Ctrl+V, with my keyboard in Greek where interpreted as Ctrl+ψ and Ctrl+ω, respectively by Thonny, just like with IDLE.

 In windows applications such Ctrl+[letter] keyboard shortcuts work flawlessly no matter what the keyboard language is set to.

Could there be some config file to add to the Cut, Copy, Paste, Select All, Undo, Redo, Find  Replace operations respective shortcuts, as they now fail with the combinations Ctrl+[non-latin Letter]?

PR #3587, opened 2025-04-03
keybord is now working  #3547

shubhamdagar9854:
Fixed an issue where keyboard shortcuts (Ctrl+C, Ctrl+X, Ctrl+V) were not recognized when using the Hebrew keyboard layout in Thonny. Now, shortcuts work regardless of the active keyboard layout. 🚀


shubhamdagar9854:
#3587

WhitespaceQ:
I think the next logical step is to extend the `*_config_page.py` with `shortcuts_config_page.py` as per #422

aivarannamaa:
Thanks for the suggestion! I'll postpone this change for now, hoping that the newer Tk version in Thonny 5.0 will fix this problem by itself.

PR #3596, opened 2025-04-12
Adds support for kitty and wezterm terminal emulators

r0zh:
Thanks for the IDE! We are using it at uni.

Just adds support for kitty and wezterm terminal emulators, which require different command formatting.


aivarannamaa:
Thank you!

PR #3716, opened 2025-10-27
docs: added Linux installation steps for Thonny

yeshuawm999:
i have included the installation steps for linux users(beginers )in this repo and my instructions are simple,they just have to copy paste this commands in their own terminal.


aivarannamaa:
Thanks for the PR, but this makes the Linux story seem simpler than it is. `apt` only works in some Linux distributions and would install an old version of Thonny in Debian and Ubuntu, so I'd rather not promote it.

PR #3824, opened 2026-03-24
Fix LSP integration

mrexodia:
While working on #2639 I found a bunch of problems with the LSP integration and Thonny hanging/crashing.

This branch is what I based my other fixes on, because testing the editor improvements is impossible without it.

Note that there is an unfixed bug/problem where opening a file in your HOME will time out all LSP requests because the LSP will attempt to index all site-packages etc and crash/hang. Thonny should be resilient to this case, but it is not fixed. I was testing in a sub folder of my HOME.

### DIFF
--- a/thonny/editors.py
+++ b/thonny/editors.py
@@ -859,6 +859,9 @@ class Editor(BaseEditor):
             )
 
         self._unpublished_incremental_changes = []
+        self._content_at_server = self.get_content()
+        self._last_fully_published_version = version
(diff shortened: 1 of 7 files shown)


aivarannamaa:
This PR contains many changes which I can tell are not necessary or which I don't see the need for (but can't say for sure). I'll pay attention to the list of the assumed problems taken from the commit messages, but I can't use the PR as is.

I'll do some bugfixing on my own and I hope the number of remaining bugs will be small enough so it is possible to report them one by one together with steps for reproduction.

PR #3833, opened 2026-04-05
Add UI-enhancement, open <untitled> tab when tab bar is clicked

aquazod:
This PR adds UI-enhancements for when the tab bar is double-clicked, a new <untitled> tab opens (VS code style)

Implementation:
- Defined the double click behavior for the tab bar, by defining the **Tkinter** event.
- Defined the double click method in the CustomNotebook class to be override by the subclasses.
- Override the double click method at the EditorNotebook subclass to open a new file, utilizing _cmd_new_file() method.

Tested on Windows.


aivarannamaa:
Looks nice! Thank you!

Issue #3818, opened 2026-03-15
Right click in Broswer doesn't make sense


Coxinha2K:
Hello there. I'm using Thonny 4.1.7 and while managing my files, i found a issue.
In Thonny's browser, if you right click in a blank space, you get a context menu that is the same of a file.
If you click something like "Open in Thonny", it gives you a error complaining about path being None, like this:
```python
Traceback (most recent call last):
  File "C:\Program Files\Python39\lib\tkinter\__init__.py", line 1885, in __call__
    return self.func(*args)
  File "C:\Users\wende\AppData\Roaming\Python\Python39\site-packages\thonny\base_file_browser.py", line 661, in <lambda>
    label=tr("Open in Thonny"), command=lambda: self.open_file(selected_path)
  File "C:\Users\wende\AppData\Roaming\Python\Python39\site-packages\thonny\base_file_browser.py", line 1075, in open_file
    get_workbench().get_editor_notebook().show_file(path)
  File "C:\Users\wende\AppData\Roaming\Python\Python39\site-packages\thonny\editors.py", line 1050, in show_file
    editor = self.get_editor(filename, True)
  File "C:\Users\wende\AppData\Roaming\Python\Python39\site-packages\thonny\editors.py", line 1131, in get_editor
    if os.path.isfile(filename_or_id):
  File "C:\Program Files\Python39\lib\genericpath.py", line 30, in isfile
    st = os.stat(path)
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```
Here is a guide:
- Enable Browser in View menu
<img width="306" height="491" alt="Image" src="https://github.com/user-attachments/assets/b49d1a74-e276-43c7-aa17-f52f40c385e9" />

- Right click in a blank space of the Browser
<img width="203" height="497" alt="Image" src="https://github.com/user-attachments/assets/f381ad33-5445-4ca5-b920-c821fb0ae8db" />

- Select a option that only a file could have, like "Open in Thonny"
<img width="275" height="589" alt="Image" src="https://github.com/user-attachments/assets/926be015-1b1d-4244-a100-1b3f3c370a01" />

- See the error in front of your eyes
<img width="921" height="440" alt="Image" src="https://github.com/user-attachments/assets/a635d0a6-a4c4-4335-8f41-b98e43e2b31e" />

aivarannamaa:
Thanks for the report! I'll investigate this later.

AgrimGupta04:
[...]

PR #3837, opened 2026-04-12
Fix: crash when right clicking blank space in file browser.

AgrimGupta04:
This PR solves the issue stated in #3818 
It ensures that actions requiring a file path are only available when an item is actually selected.
1. Wrapped Open, Cut, Copy, Rename, Delete, and Properties in a check for selected_node_id. These items no longer appear when right-clicking the background.
2. Maintained the "Paste" functionality on blank space by falling back to the current directory (`current_focus` or `active_directory`), allowing users to paste files into the open folder without selecting a sub-folder first.

### DIFF
--- a/thonny/base_file_browser.py
+++ b/thonny/base_file_browser.py
@@ -681,9 +681,13 @@ class BaseFileBrowser(ttk.Frame):
         return "break"
 
     def open_file(self, path):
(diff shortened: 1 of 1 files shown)


aivarannamaa:
Thank you!

Issue #3832, opened 2026-04-04
Add the feature about really saving text into the clipboard


SkyElysium:
My System: Windows 10

If I copy some text in Thonny and then exit, I can't paste it beacuse the text in clipboard has been cleaned.
(It looks like Tkinter doesn't really save it into the clipboard)

It's not convenient. I hope the text can be really saved into the clipboard.

AgrimGupta04:
Hi ! I've investigated this issue and have a working fix that resolves it strictly within workbench.py.

The Root Cause:
On Windows, Tkinter's shutdown by clicking the 'X' button runs a cleanup sequence that clears the clipboard ownership held by the Tcl interpreter. I confirmed this by noticing that the clipboard persists if the process is terminated abruptly (Ctrl+C), but is wiped during a normal exit.

The Fix:
I've cleaned up the exit(destroy) logic in `workbench.py` to ensure the existing clipboard preservation code is always executed before the Tcl interpreter is shut down. 
This approach works with the existing code in `misc_utils.py` and `ui_utils.py` and requires no new dependencies. I've tested it on Windows 11 and it fixes the issue for both the editor and the file browser.

May I raise a Pull Request for this?

SkyElysium:
Thanks for solving this issue. I think you could do that. And it's also helpful for my project to solve the same problem.

aivarannamaa:
@AgrimGupta04, please make the PR, as I'm not sure I understood your idea from the description only.

aivarannamaa:
I could confirm that the old approach does not work anymore, but I found another recommendation (https://github.com/python/cpython/issues/84632#issuecomment-2379692208) which seems to work. I'll try to releases 5.0.0 tonight, so you can test it yourself.

PR #3845, opened 2026-04-21
Fix: clipboard clearing on exit

AgrimGupta04:
This PR solves the issue stated in #3832
It makes sure that the clipboard holds the text in it and also able to paste the same post exiting thonny.
The changes include:

1. Early Clipboard Sync: Added a call to `self.clipboard_get()` at the very beginning of the `destroy()` method. This forces a synchronization between Thonny and the Windows System Clipboard while the UI thread is still fully active. 
2. Performance Gatekeeping: Added a length check (> 1000 characters). If the text is large, we assume it is code/text and skip the disk-check logic entirely.

### DIFF
--- a/thonny/base_file_browser.py
+++ b/thonny/base_file_browser.py
@@ -681,9 +681,13 @@ class BaseFileBrowser(ttk.Frame):
         return "break"
 
     def open_file(self, path):
+        if not path:
+            return 
         pass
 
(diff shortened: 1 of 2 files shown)


aivarannamaa:
In your PR you are basically just omitting the call to `copy_to_clipboard` and according to my tests, this doesn't fix the problem. I found another solution though.

PR #3879, opened 2026-05-14
docs: fix spelling eg. -> e.g.

### LINKED ISSUE #3869: Correct "eg." spelling error in docs
Hi,
I have noticed that there is a slight spelling error in some of the wiki pages, as you have written "eg." instead of the expected spelling "e.g.", which seperates the 2 letters from each other. Normally, Latin expressions have one dot, but for this one, and i.e.., they have 2 apparently.

Here is a patch you can apply on the wiki to fix this small problem across 9 files, generated by just doing search-and-replace.

[0001-chore-docs-Fix-wrong-spelling-of-e.g.-eg.-in-docs.patch](https://github.com/user-attachments/files/27304799/0001-chore-docs-Fix-wrong-spelling-of-e.g.-eg.-in-docs.patch)

linbo328:
Fixed: eg. -> e.g. in translate.md

Closes #3869


Sampo-Kolu:
Thanks, I only searched the typos in the docs repo in the patch I made for #3869, so I would have missed this because it's in the main repo. Thanks for detecting this. However, the patch in #3869 must be applied to the docs repo as well to fully close it.

